PR #6, opened 2024-02-29
feat(Twitter): Hide Recommended Users, Hide For You

Swakshan:
Better way of hiding recommended users. (instead of JsonHooks) and For you from timeline


bqv:
Can this be done for the Explore page too - remove the "For you" trends tab, or just default to the next ("Trends") tab?

bqv:
@crimera

PR #29, opened 2024-03-16
Update README.md

fuwamocoanon:
Added descriptions for new patches that were missing.


fuwamocoanon:
Obviously only needed once done. Figured these would be the best explanations.

crimera:
> Obviously only needed once done. Figured these would be the best explanations.

All the patches listed in the readme are already done.

PR #55, opened 2024-03-29
feat(ui): bring back old twitter strings

IndusAryan:
photo_2_2024-03-29_18-34-17

photo_1_2024-03-29_18-34-17

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,3 +31,4 @@ java -jar cli.jar patch \
 | `Hide Banner` | Hide new post banner |
 | `Hide Live Threads` | Hides the Live Threads feature from main timeline|
 | `Enable custom app icon and nav icon settings` | Unlock custom nav bars and app icons |
+| `Bring back Twitter` | Bring back old Twitter logo and name |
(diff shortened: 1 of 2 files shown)


crimera (commented):
I think we can just iterate through all the strings and replace all instances of post to tweet.

IndusAryan:
the map approach may look a bit manual but it was so because i wanted to change only words that are more visible to users, not everywhere in the app as it can break stuff

crimera:
Cool

> the map key value pair approach may look a bit manual but it was so because i wanted to change only words that are visible to users, not everywhere in the app as it can break stuff plus suppose if we add more strings there then it will reach 50 lines max as only mainstream keywords are important and noticeable

cool, yeah think we should go for stability.

crimera:
> otherwise this one can change these 4 words
> 
> ```kotlin
> private fun updateStringsFile(stringsFile: File, context: ResourceContext) {
>         context.xmlEditor[stringsFile.toString()].use { editor ->
>             val document = editor.file
> 
>             val nodes = document.getElementsByTagName("string")
> 
>             for (i in 0 until nodes.length) {
>                 val node = nodes.item(i)
>                 var content = node.textContent
> 
>                 /** Word boundary (\\b) regex pattern keeps casing of chars intact,
>                  *  i.e, post to tweet && Post to Tweet **/
> 
>                 content = content.replace("\\bpost\\b".toRegex(), "tweet")
>                 content = content.replace("\\brepost\\b".toRegex(), "retweet")
>                 content = content.replace("\\bposted\\b".toRegex(), "tweeted")
>                 content = content.replace("\\breposted\\b".toRegex(), "retweeted")
>                 content = content.replace("\\bposts\\b".toRegex(), "retweets")
> 
>                 // Update the node content if changes were made
>                 if (content != node.textContent) {
>                     node.textContent = content
>                 }
>             }
>         }
>     }
>     ```
> ```

but I like looking at this more concise code though 😆

PR #57, opened 2024-03-30
feat(twitter): add more strings and move strings map to a different object

IndusAryan:
i think i have translated enough strings that are noticiable to users.
but i am unsure about this ,will this stop patching of this patch midway ??? if any key is not found, probably not afaik.
```kotlin
if (!keyReplaced) {
                    //println("Key not found: $key")
                    throw PatchException("key not found: $key")


crimera:
> will this stop patching of this patch midway ??? if any key is not found

It will

IndusAryan:
rolled it back

PR #109, opened 2024-04-15
feat(ui): full app wide material theming in dim mode replacing dark blue

IndusAryan:
few screenshots:
https://github.com/IndusAryan/twitter-patches?tab=readme-ov-file#screenshots


Mehlsine:
Is this going to be a separate patch or a replacement for the current one? I believe some of us would prefer dynamic colors solely for UI Elements (without colorful background) as it is right now.

IndusAryan:
will they like same blue theme over material one 🤔

crimera:
> Is this going to be a separate patch or a replacement for the current one? I believe some of us would prefer dynamic colors solely for UI Elements (without colorful background) as they are right now.

The patch will only apply the theme on dim mode. Is that a theme you use?

PR #130, opened 2024-04-20
Update Russian translation

Mehlsine:


IndusAryan:
💀 you dont have to add these, strings are by default used from global file

Mehlsine:
💀

PR #211, opened 2024-05-16
feat(ui): app wide monet theming in light mode

IndusAryan:
photo_1_2024-05-16_13-56-20

photo_2_2024-05-16_13-56-20

my app is not logging in for some reason after changing package and i cant uninstall og for some reason, please share screenshot while testing of home screen, i think the divider color 'may' require some change.


crimera:
Screenshot_20240517-235057_X

Looks good

PR #212, opened 2024-05-16
refactor: update hindi translations, change root project name & lowercase fab package

IndusAryan:


indrastorms:
It should not include "twitter" in root project name as there are upcoming Instagram patch and maybe more in the future.

PR #233, opened 2024-05-21
fix: URL decode path to JAR containing spaces by osumatrix

taku-nm:
[this fix ](https://github.com/ReVanced/revanced-patches/pull/3079) by osumatrix should also be applied here


kitadai31:
Also use this fix
https://github.com/ReVanced/revanced-patches/commit/d42fbb152126cf2177315c4706fb03bc89f5af1c
That version causes error on below Android 12

PR #263, opened 2024-06-04
Translations: Update Brazilian Portuguese

r7kings:


kitadai31:
it seems wrong file is edited

PR #310, opened 2024-07-14
The X icon changes to the Twitter icon.

lgernierO:


lgernierO:
@crimera Hello, is there any problem with this one? I saw that there is an X icon in the mipmap-hdpi of bringbacktwitter, and the twitter icon is everywhere else. I replaced it with the twitter icon.

crimera:
Oh ok, sorry I thought the commit only deleted the icon. My bad.

PR #342, opened 2024-08-12
Get back twitter home icon

monefiera:
This is a pull request to change the Twitter home icon back to the original birdcage.
Please let me know if there are any elements that should be added.


IndusAryan (changes requested):
do ctrl + alt + L in each xml, the xml vector syntax is very fucky and can literally crash the app sometimes

Issue #442, opened 2024-10-22
Fingerprint for "Open browser chooser on opening links" is outdated


dic1911:
As title, the old wrapper method is not there anymore, attached screenshot shows currently working patch

Matching pattern:
 - signature: `(Landroid/content/Context;Ljava/lang/String;)Z`
 - uses string `android.intent.category.BROWSABLE`

(2 commented blocks with same smali code)
![image](https://github.com/user-attachments/assets/d6da83b1-a839-48cc-aee3-b9fbac2e7b47)

```
    # increase local register count to 8 first
    new-instance v6, Landroid/content/Intent;
    const-string v7, "android.intent.action.VIEW"
    invoke-virtual {v3}, Landroid/content/Intent;->getData()Landroid/net/Uri;
    move-result-object p2
    invoke-direct {v6, v7, p2}, Landroid/content/Intent;-><init>(Ljava/lang/String;Landroid/net/Uri;)V
    const v7, 0x0
    invoke-static {v6, v7}, Landroid/content/Intent;->createChooser(Landroid/content/Intent;Ljava/lang/CharSequence;)Landroid/content/Intent;
    move-result-object v3
```

dic1911:
I suppose that we're done here.

PR #540, opened 2025-03-09
Some improvements

### LINKED ISSUE #422: bug(Open browser chooser on opening links): Setting description is misleading
https://github.com/crimera/piko/blob/82e912a25687c891ae61c97a0d414168718b6fc8/src/main/resources/twitter/settings/values/strings.xml#L62

This description says "instead of using the in-app browser", but disabling the in-app browser can be achieved with an official Twitter setting.
By this description, some users seem to misunderstand that this settings is for disabling the in-app browser.

I suggest this description.

```diff
-    <string name="piko_pref_browser_chooser_desc">Open links with browser chooser instead of using the in-app browser</string>
+    <string name="piko_pref_browser_chooser_desc">Open links with browser chooser instead of one of the installed browsers</string>
```

This suggestion is partially taken from the patch description, which describes more correctly the how the patch works:

https://github.com/crimera/piko/blob/82e912a25687c891ae61c97a0d414168718b6fc8/src/main/kotlin/crimera/patches/twitter/link/browserchooser/BrowserChooserPatch.kt#L12-L17

### LINKED ISSUE #492: Since 10.68 not working manager rvx error
- Device Info
RVX Manager: 1.22.2
Model: VNE-LX3
Android version: 14
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Root permissions: No

- Patch Info
App: com.twitter.android v10.70.0-release.0 (Suggested: Any)
Patches version: v1.46.1
Patches added: Bring back twitter, Enable app downgrading, Export all activities
Patches removed: Dynamic color, Remove Detailed posts, Remove "Pinned posts by followers" Banner, Remove message prompts Banner, Remove "Revisit Bookmark" Banner, Remove videos for you, Remove "Who to follow" Banner, Hide bookmark icon in timeline, Hide FAB Menu Buttons, Hide immersive player, Remove view count, Hide FAB, Hide Recommended Users, Hide Banner, Hide hidden replies, Hide nudge button, Hide followed by context, Hide Live Threads, Show poll results, Hide promote button
Default patch options changed: None

- Settings
Allow changing patch selection: true
Version compatibility check: true
Show universal patches: true
Remove unused library: false
Use pre-release: true
Patches source: crimera/piko
Integration source: crimera/revanced-integrations

- Logs
Reading APK
Decoding app manifest
Loading patches
Merging integrations
Deleting existing temporary files directory
Decoding resources
Executing patches
Applied 44 patches
Compiling patched dex files
Compiled 12 dex files
Compiling modified resources
An error occurred:
brut.androlib.exceptions.AndrolibException: brut.common.BrutException: could not exec (exit code = 1): [/data/app/~~v57Zvhx1Ux0hWpiAZmUIOw==/app.rvx.manager.flutter-xeW-j6rTcuFMwaLZtBdfig==/lib/arm64/libaapt2.so, compile, --dir, /data/user/0/app.rvx.manager.flutter/files/patcher/tmp-CUEPSM/revanced-temporary-files/apk/res, --legacy, -o, /data/user/0/app.rvx.manager.flutter/files/patcher/tmp-CUEPSM/revanced-temporary-files/apk/build/resources.zip]
	at brut.androlib.AaptInvoker.invokeAapt2(SourceFile:123)
	at brut.androlib.AaptInvoker.invokeAapt(SourceFile:394)
	at app.revanced.patcher.data.ResourceContext.get(SourceFile:134)
	at app.revanced.patcher.Patcher.get(SourceFile:268)
	at app.revanced.manager.flutter.MainActivity.runPatcher$lambda$35(SourceFile:343)
	at app.revanced.manager.flutter.MainActivity.$r8$lambda$7KobHryxqqdI5JS3uiaGATjlYU8(SourceFile:0)
	at app.revanced.manager.flutter.MainActivity$$ExternalSyntheticLambda2.run(SourceFile:0)
	at java.lang.Thread.run(Thread.java:1012)
Caused by: brut.common.BrutException: could not exec (exit code = 1): [/data/app/~~v57Zvhx1Ux0hWpiAZmUIOw==/app.rvx.manager.flutter-xeW-j6rTcuFMwaLZtBdfig==/lib/arm64/libaapt2.so, compile, --dir, /data/user/0/app.rvx.manager.flutter/files/patcher/tmp-CUEPSM/revanced-temporary-files/apk/res, --legacy, -o, /data/user/0/app.rvx.manager.flutter/files/patcher/tmp-CUEPSM/revanced-temporary-files/apk/build/resources.zip]
	at brut.util.OS.exec(SourceFile:104)
	at brut.androlib.AaptInvoker.invokeAapt2(SourceFile:119)
	... 7 more
Something went wrong:
PathNotFoundException: Cannot copy file to '/data/user/0/app.rvx.manager.flutter/files/lastPatchedApp.apk', path = '/data/user/0/app.rvx.manager.flutter/files/patcher/tmp-CUEPSM/out.apk' (OS Error: No such file or directory, errno = 2)

### LINKED ISSUE #500: How do I install this thing?
`Set crimera as integration organization. ` doesn't even show up on Revance Mangaer as a field I can set up. And the Termux thingy, revancify, will never patch, it either takes an hour to download 10MB of data or throws an error every time.

image

### LINKED ISSUE #508: Use Browser Chooser isn't taking effect
Using release x-piko-v10.72.2-release.0.apk from 
the twitter apk repository
I've verified that Use Browser Chooser patch is included

Option Misc/Use Browser Chooser has no effect. On or off, links open in the in-app Browser

kitadai31:
Hi 👋
I made some improvements of piko patches.
This pull request contains 6 commits.

#### 1. fix(Bring back twitter): Resource compilation fails on ReVanced Manager

- Fixes #492

#### 2. fix(Enable Reader Mode): Specify the compatible version, show a warning instead of throwing an exception when failed

Compatible version is useful for users who want to use reader mode.
Also, if the patch has failed, show a warning instead of showing a wall of the stack trace.

`WARNING: "Enable Reader Mode" is not supported in this version. Use X 10.72.2 or earlier. The patch was not executed.`

[Screenshot](https://github.com/user-attachments/assets/6901f5ef-2011-4d11-8ca5-3fe9c5ca23b0)

#### 3. docs: Add compatible CLI and Manager versions to README until bumping to ReVanced Patcher v21

- Fixes #500

This is absolutely necessary!

#### 4. fix(Remove Detailed posts): Change the settings label from "detailed" to "related"

"Tweet Detail Related Tweets" was mistakenly abbreviated to "Detailed".
"Related" is more accurate word to describe "Discover more" section.

Note: The name of the patch is not changed in this commit.
Because renaming the patch name might affect in manager/builder that's use this patch name.
Variable names and string keys are also not changed as per the patch name.
Only the setting label has been changed.

I also fixed a typo "Revist" as well.

#### 5. fix: Remove non-functional `Open browser chooser on opening links` patch

The last working version of this patch was `10.48.0-release.0`.
So, I tried updating this patch according to #442.
Then, I fixed the patch and I checked that the integrations method (`openWithChooser`) is called correctly when I open any links. (Checked with logs)

However, the app chooser still doesn't open.
It seems that `Intent.createChooser()` has no effect.
Tested on Pixel 3, Android 14 (LineageOS 21).
3 browsers are installed: Chrome, Firefox and Via.

So I think this patch is not working and should be removed.

I also tested the link which has an external app.
For example, I installed Pixiv app on my phone, so when I opened a Pixiv link on patched Twitter, the app chooser should be opened and show "Chrome or Firefox or Via or Pixiv?" choices.
But the app chooser doesn't open in this situation also.

For reference, these are the patch source I used.
https://github.com/kitadai31/piko/blob/5d1e295c41f9b98299e1de820bd7d7ba5d90de7a/src/main/kotlin/crimera/patches/twitter/link/browserchooser/BrowserChooserPatch.kt
https://github.com/kitadai31/piko/blob/5d1e295c41f9b98299e1de820bd7d7ba5d90de7a/src/main/kotlin/crimera/patches/twitter/link/browserchooser/fingerprints/OpenLinkFingerprint.kt

> @dic1911
> Hello, if you are reading this, can you please let me know if you actually managed to open the app chooser by #442 method?
> And what Android version are you using?
> I suspect this has to do with a change in Android 12.

- Closes #422
- Closes #508 

#### 6. feat(Settings): Add a description to "Native features" page

The usage of native features were unclear.
This helps users find the native features.

<details>

<summary>Screenshot</summary>

</details>

I have one question for this commit.

I changed "detailed" to "related" only for English strings.
I checked with Google Translate and found that it is still translated as "detailed post" in some other languages.
Should I delete the changed string from all languages?
Or shouldn't?


dic1911:
> Hello, if you are reading this, can you please let me know if you actually managed to open the app chooser by https://github.com/crimera/piko/issues/442 method?
And what Android version are you using?
I suspect this has to do with a change in Android 12

Hi, thanks for trying to help, I was on Android 14 back then and now on 15, the provided patch is working on both Android version for Twitter 10.63 (Sorry but I don't really update modded apks unless something breaks so I couldn't verify if the same applies to newer updates)

kitadai31:
Oh, really? Hmm... 
Then it may be a target Twitter version problem? I patched 10.82.0.

> edit: @ kitadai31 btw if you want I can share my apk so you can easily test it

Thank you 🙏
I want to test with it
If you have a Telegram or Discord, please send to @ kitadai31

mrndstvndv:
These patches are much appreciated. Thank you

> Should I delete the changed string from all languages?
Or shouldn't?

nah lets keep them.

dic1911:
@kitadai31 since we're here, can you also send me your patched APK to check if the smali instructions were patched properly?

kitadai31:
@dic1911 I tried your manually modified 10.63.1 apk, but the app chooser still don't open.

However, just only once, when I tapped on an external link on a profile page, the app chooser appeared, but it said "There is no app for this action."
After that, the app chooser never appeared again. I can't reproduce it. 😕

OK, I'll send my patched apk tomorrow

dic1911:
> @dic1911 I tried your manually modified 10.63.1 apk, but the app chooser still don't open.
> 
> However, just only once, when I tapped on an external link on a profile page, the app chooser appeared, but it said "There is no app for this action." After that, the app chooser never appeared again. I can't reproduce it. 😕

This sound like what would happen when you have default browser set, can you try to freeze the browser which it opens automatically when you tap on links?

kitadai31:
@dic1911 

> can you try to freeze the browser which it opens automatically when you tap on links?

It works!
After I unset the default browser by uninstalling it, the browser chooser started to appear.
Working on both your 10.63 apk and my patched 10.82 apk.

Screenshot_20250311-142600_IntentResolver

I thought that patch allows us to choose browser every time even if the default browser is set.
It might be better to add a note about it in the description of the "Use browser chooser" setting.

kitadai31:
But, if the default browser is not set, the stock Twitter also opens a normal intent chooser.
(You need to turn off X settings > Display > "Use in-app browser")

Screenshot_20250311-163649_Android System

This dialog is a little different from the app chooser in that it has an "Always" button, but I think it is almost same.
So, is this patch necessary...?

dic1911:
@kitadai31 
> even if the default browser is set.

It's just how Android works, AFAIK when a default browser is set and you want to open the link with a chooser for something else, you need to iterate through apps to check what apps can open the link and make UI yourself

kitadai31:
@dic1911 How about this? https://github.com/crimera/piko/pull/540#issuecomment-2713391106
I found that the stock Twitter also opens browser chooser if the default browser is not set.
Therefore, this patch may no longer be needed.

It is possible that the behavior of the Twitter app has changed in recent versions after you contributed the browser chooser patch.

But if you think that patch is still useful, of course I'll drop the commit d654ba462a0e7b01bba4a10d4589e8b73749e813 and fix the patch instead.

dic1911:
@kitadai31 Since I don't really know when it's changed, maybe just leave the patch there and add a note so people would know it should only be applied on older versions?

kitadai31:
> It is possible that the behavior of the Twitter app has changed in recent versions after you contributed the browser chooser patch.

↑ To see if it is correct, I just tried Twitter 10.23.0.
As a result, this version already opens the browser selector if no default browser is set.
Tested on Android 7.1 and 14.
So this was not a change in recent versions.
Maybe it was an oversight.

> Since I don't really know when it's changed, maybe just leave the patch there and add a note so people would know it should only be applied on older versions?

Based on the above result, I don't think that's necessary.
10.23.0 is a sufficiently old version.
And it's too old to apply the piko patch.
(None of the patches are available for 10.23.0 because the build fails due to `Settings` patch.)
Therefore, the browser selector should appear even on the oldest Twitter that can apply the piko patch.

@dic1911
For these reasons, may I remove the patch?

> Alternatively, I can fix the fingerprint so that the patch fails successfully in 10.49+.
> The current fingerprint incorrectly matches the Integrations method `openWithChooser`, which causes the patch to succeed on 10.49+ accidentally.

dic1911:
@kitadai31 I'd ask @crimera for opinion instead since I don't really own piko ;)

kitadai31:
> I'd ask crimera for opinion instead since I don't really own piko ;)

Fair enough.

I re-pushed the PR branch with some updates:

- Simplified the check for Manager in Bring back twitter patch
- Changed the commit message for the browser chooser patch commit

@crimera This PR is ready for review!

PR #556, opened 2025-04-09
feat(Translations): Korean translation added

BlackGold8282:
feat(Translations): Add Korean


BlackGold8282:
Do I need to do anything else besides modifying the XML file?

PR #674, opened 2025-09-24
update(Translations): Update `Polish`

krvstek:


kitadai31:
Are the Bring back twitter strings correct?

I reference the strings in Twitter 9.98.0.
`tweets_retweet` is `Podaj dalej`
`tweets_retweeted` is `%s podał/a dalej Tweeta`

As far as I know, these strings do not necessarily need to match the 9.98.0 version exactly, but I think it is better to use the official names in your language for proper nouns like “Tweet” and “Retweet” so that users can actually go back to old twitter.

Also, you must add a backslash before double quotes in strings.xml
Same as single quotes

krvstek:
> Are the Bring back twitter strings correct?
> 
> I reference the strings in Twitter 9.98.0. `tweets_retweet` is `Podaj dalej` `tweets_retweeted` is `%s podał/a dalej Tweeta`
> 
> As far as I know, these strings do not necessarily need to match the 9.98.0 version exactly, but I think it is better to use the official names in your language for proper nouns like “Tweet” and “Retweet” so that users can actually go back to old twitter.
> 
> Also, you must add a backslash before double quotes in strings.xml Same as single quotes

I didn't take the official names, when adding this translation to the patch, as I had no idea those terms were actually translated in the original app. As for the backslashes: I just followed the original translation and from what I can see, there are no backslashes before quotation marks in the file.

I'll apply a fix for this as soon as I can.

PR #683, opened 2025-09-29
Some more fixes for v21

swakwork:


patches/src/main/kotlin/app/crimera/patches/twitter/misc/bringbacktwitter/BringBackTwitterPatch.kt
```diff
@@ -153,17 +164,19 @@ val bringBackTwitterPatch =
...
+                    replaceStringsInFile(
+                        it,
+                        replacements =
+                            mapOf(
+                                "X" to "Twitter",
+                                "ポスト" to "ツイート",
+                            ),
+                    )
+                }
             }
```
kitadai31, line 180: String replacement for Japanese is only run when `isRunningOnManager`

kitadai31: Also, `setOf` ~ `forEach` is no longer needed

swakwork: lol I thought, isRunningOnManager is required here as well, will correct it, thanks

PR #708, opened 2025-10-19
updates

kitadai31:
- refactor(Dynamic color): Do not get elements repeatedly in XML operation

See commit message

- feat(Change version code): Add compatibleWith to make it a non-universal patch

In the crimera's uncompleted patcher v20 branch, crimera had added `compatibleWith("com.twitter.android")` to the Change version code patch.
https://github.com/crimera/piko/pull/354/files#diff-a884a22f65e4820fa6514a09be3f37a2a763f981fc9bd6b40a1fe49ab30df205R12
I like this idea, so I think it should be reflected to v21 branch too.
Since piko patches is only targeting for Twitter, universal patches are no point.
There is no problem adding compatibleWith for universal patches.
If piko add support for other app like Grok in the future, it is enough to just add that package name to the compatibleWith.

However, this PR did not touch another universal patch, the `Export all activities`.
Because I felt that this patch would have a negative impact if it were to appear in Manager by default.
(But in the v20 branch, this patch had also became a non-universal patch.)
If this patch also should be non-universal, I'll add compatibleWith.

- ~~update: Remove `Delete from database` patch and make it a dependency of other patches~~

This patch was like a "dummy" since it only unlocked settings.
Rather than making it a standalone patch, I made it as a dependency for other patches that remove the timeline entry.
With this PR, if users have applied either the “Remove Ads” patch or the “Hide timeline posts by category” patch, the “Remove entries from database” option will appear in the Ads settings, which will be convenient.

- fix(Custom font): Fix crash on Android 8 by restricting the functionality to Android 9+

Subject says it all

- feat(Settings): Allow preference titles to be multiple lines

Previously, preference titles were limited to one line, which resulted in titles being truncated depending on the language and user screen size settings.
This was especially problematic in my native language, where character limits were strict.
This fixes truncation of titles.

Screenshot_20251020-023045_Twitter

Screenshot_20251020-023348_X

Some of changes may need to be discussed. e.g. universal patch change and remove of database patch
Please check when you have time.

### DIFF
--- a/extensions/twitter/src/main/java/app/revanced/extension/twitter/settings/widgets/Helper.java
+++ b/extensions/twitter/src/main/java/app/revanced/extension/twitter/settings/widgets/Helper.java
@@ -28,12 +28,14 @@ public class Helper {
         preference.setSummary(summary);
         preference.setKey(setting.key);
         preference.setDefaultValue(setting.defaultValue);
+        preference.setSingleLineTitle(false);
         return preference;
     }
 
     public Preference editTextNumPreference(String title, String summary, StringSetting setting) {
         EditTextPref preference = (EditTextPref)editTextPreference(title,summary,setting);
         preference.setNumericOnly(true);
+        preference.setSingleLineTitle(false);
         return preference;
     }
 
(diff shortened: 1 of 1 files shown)


swakwork:
> but then the question arises: which page should it be on? 🤔

How about `Misc` ?

PR #726, opened 2025-11-08
Support legacy embed domains for vxTwitter

aubymori:
This PR adds the `l.vxtwitter.com` and `old.vxtwitter.com` domains (which return to a legacy method of embedding on Discord) to the custom deeplinks list.


kitadai31:
In the dev branch, all subdomains are added automatically

PR #751, opened 2025-12-02
Update(Traslations): Update Italian 

speedlorenzo12:


swakwork:
Merged.

Going forward set the commit message as _" Update(Translations): Update `language-name` "_ if possible, so that it gets reflected in changelogs.

Thanks,

PR #794, opened 2026-02-14
Morphe compatibility

wchill:


Parsa307:
@LisoUseInAIKyrios, any plans to add this to the morphe patches repo?

wchill:
My understanding is that Piko will remain separate, but we will ship it as an available-by-default patch bundle in Morphe Manager.

PR #801, opened 2026-02-19
refactor(Twitter): Rename `Hide timeline posts by category` patch to `Hide recommendation items`

kitadai31:
I've always thought that the name of "Hide timeline posts by category" patch is confusing since we migrated to patcher v21.
This patch name could be misunderstood to mean that it hides timeline tweets based on categories (≒ genres) such as “sports,” “politics,” or “entertainment.”

Additionally, this patch actually hides more than just posts.
For example, the "Who to follow" section, which this patch hides, is not a "post".

I suggest the name "Hide timeline items".
I think the word "item" describes both posts and sections well.

What do you think?

This PR is marked as draft because I think this change requires discussion.


swakwork:
Could you rebase this as well.

kitadai31:
Done.

PR #814, opened 2026-02-25
refactor: Use AddResourcePatch to standardize ARSCLib strings usage

LisoUseInAIKyrios:
Changes string/array resources to use `AddResourcesPatch` that modifies the target app files.

This is required for ARSCLib to process the files, as it doesn't handle adding files like Apktool does.

### DIFF
--- a/patches/src/main/kotlin/app/morphe/patches/all/misc/resources/AddResourcesPatch.kt
+++ b/patches/src/main/kotlin/app/morphe/patches/all/misc/resources/AddResourcesPatch.kt
@@ -43,6 +43,8 @@ import app.morphe.patches.all.misc.resources.StringResourceSanitizer.sanitizeAnd
 import app.morphe.util.forEachChildElement
 import app.morphe.util.getNode
 import app.morphe.util.inputStreamFromBundledResource
+import org.w3c.dom.Element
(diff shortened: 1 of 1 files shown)


patches/src/main/kotlin/app/morphe/patches/all/misc/resources/AddResourcesPatch.kt
```diff
@@ -0,0 +1,276 @@
...
+import java.util.Locale
+import java.util.logging.Level
+import java.util.logging.Logger
+
+internal val locales = listOf(
+    AppLocale("", ""), // Default English locale. Must be first.
+    AppLocale("ar-rSA", "ar"),
+    AppLocale("es-rES", "es"),
+    AppLocale("fr-rFR", "fr"),
+//    AppLocale("hi-rIN", "hi"), // FIXME. Uncomment after patcher fixes strings.xml after decoding
```
LisoUseInAIKyrios: The unpatched X Hindi strings.xml has a typo of a non valid XML character and it breaks the add resource patch when it tries to load the invalid strings.xml.

This will be fixed in patcher by fixing XML errors after decoding instead of before encoding.

After patcher is fixed this commented out line should be restored.

PR #958, opened 2026-04-11
feat(Instagram): Disable read receipts in DMs

lewislarsen:
This PR is intended to prevent read receipts (i.e. Seen 5 minutes ago) from being sent when opening Instagram DMs. 

The endpoints may need to be adjusted, as the only sources I could find are quite old and may not represent the true endpoints used by Instagram in 2026.


swakwork:
Thanks. Can you check one thing ?
The last time I retired to block the url, the chat notified was persistent ie, when a new message appears and when I open the chat, read it & close the chat, I was still noticed that the message was unread aka the chat was highlighted white. (Hope I'm making sense)


Also could you open the PR on dev branch not main.

PJGitHub9:
Would be nice for a "mark as read" option 😄

xchwarze:
@swakwork If I remember correctly, that used to cause the "seen" count to stop updating. It was a widely used feature in MyInsta v25/v26.

swakwork:
Included in https://github.com/crimera/piko/releases/tag/v3.3.0-dev.1

PR #961, opened 2026-04-12
feat: Anonymous DM viewing in Instagram

EncryptedCicada:
Adds an option in the settings to toggle settings DMs as read.

This patch could not be blocked by simply blocking the URI because of two main reasons:

1: The URI it uses (as far as I could intercept) is the general graphql_www endpoint and it actually encodes the request to mark as read in a field named fb_api_req_friendly_name (= IGDirectItemSeenMutation), which from what I understood from the codebase can't easily be intercepted yet.

2. I am not really experienced with working with apis and if there is a way that I am unaware of to block this specific request type then changing the patch to do that would be better and cleaner.

I have the request intercept to help make this work if it turns out to be feasible. In the meantime, this patch is tested working with v423.0.0.47.66 which is the stable supported version.

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/Strings.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/Strings.java
@@ -38,6 +38,7 @@ public class Strings {
     public static final String CATEGORY_GHOST = "Ghost";
     public static final String VIEW_STORIES_ANONYMOUSLY = "View stories anonymously";
     public static final String VIEW_LIVE_ANONYMOUSLY = "View live anonymously";
+    public static final String VIEW_DM_ANONYMOUSLY = "View direct messages anonymously";
 
     public static final String CATEGORY_DISTRACTION_FREE = "Distraction free";
     public static final String DISABLE_STORIES = "Disable stories";
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/Settings.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/Settings.java
@@ -27,6 +27,7 @@ public class Settings extends BaseSettings {
     public static final BooleanSetting DISABLE_ANALYTICS = new BooleanSetting("disable_analytics", true);
     public static final BooleanSetting VIEW_STORIES_ANONYMOUSLY = new BooleanSetting("view_stories_anonymously", false);
     public static final BooleanSetting VIEW_LIVE_ANONYMOUSLY = new BooleanSetting("view_live_anonymously", true);
+    public static final BooleanSetting VIEW_DM_ANONYMOUSLY = new BooleanSetting("view_dm_anonymously", true);
     public static final BooleanSetting DISABLE_STORIES = new BooleanSetting("disable_stories", false);
     public static final BooleanSetting HIDE_STORIES_TRAY = new BooleanSetting("hide_stories_tray", false);
     public static final BooleanSetting DISABLE_EXPLORE = new BooleanSetting("disable_explore", false);
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/SettingsStatus.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/SettingsStatus.java
@@ -59,8 +59,12 @@ public class SettingsStatus {
     public static void viewLiveAnonymously() {
         viewLiveAnonymously = true;
     }
+    public static boolean viewDmAnonymously = false;
+    public static void viewDmAnonymously() {
+        viewDmAnonymously = true;
+    }
     public static boolean ghostSection() {
-        return (viewStoriesAnonymously || viewLiveAnonymously);
+        return (viewStoriesAnonymously || viewLiveAnonymously || viewDmAnonymously);
     }
 
 
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/ScreenBuilder.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/ScreenBuilder.java
@@ -144,6 +144,15 @@ public class ScreenBuilder {
                     )
             );
         }
+        if (SettingsStatus.viewDmAnonymously) {
+            addPreference(category,
+                    helper.switchPreference(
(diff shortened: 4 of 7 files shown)


EncryptedCicada:
From what I can see there have already been two patches for this purpose so if the maintainers see fit, they should close this one.

swakwork:
@EncryptedCicada have you checked the feature at your end ? 
Like I have mentioned on https://github.com/crimera/piko/pull/958#issuecomment-4229503111 if I try to block the direct seen API, the DMs are not marked as seen, hence I get an persistent notification that the user has sent a new DM

EncryptedCicada:
@swakwork I've tested it at my end and the behavior is that the notification gets dismissed but the message remains unseen.

swakwork:
> but the message remains unseen.

Yup that's why I haven't added the feature yet

EncryptedCicada:
> Yup that's why I haven't added the feature yet

But isn't that a perfect indication of which chats have not been read? How else would the user know which chats to come back to?

swakwork:
> But isn't that a perfect indication of which chats have not been read

I haven't used the feature in any mod, but if the chats are always marked as unread, wouldn't it be annoying to see the chat always be highlighted as white, and also having a persistent indication that the chat is unread ?

EncryptedCicada:
> I haven't used the feature in any mod, but if the chats are always marked as unread, wouldn't it be annoying to see the chat always be highlighted as white, and also having a persistent indication that the chat is unread ?

I understand now what you mean to say. As for what this patch is supposed to do is make the chat not appear read to the user and the person messaging them. I would think that there is no easy way to disable just the read receipts for the person messaging the user of this patch while the user is able to mark the chat as read client side.

The various mods of Instagram that I've used have injected a custom button to the chat long press menu in the inbox named "mark as read" which does exactly what it sounds like. However, unless the user pressed that, the chat would appear unread no matter the number of times the user opens to read it.

To make a client side change, the patch would be too complicated to be stable. One way we could do is to hook the method that marks the UI changes while keeping a database of the chat/thread ids that are to be marked read based on user interaction. This will have two drawbacks.
1: Like anonymously viewing stories, the change will not be persistent across sessions.
2: The patch to do this might not be feasible or if it is feasible may not be stable enough to be maintained.

One way I would've tried to implement such a patch would be to hijack the network layer and read the outgoing request body for a specific field which is `fb_api_req_friendly_name`. When it is = `IGDirectItemSeenMutation`, we can identify, by looking at the variables field which chat is being requested to be marked read. Then we maintain a custom database where we store this id field while dropping/nullifying that request. However this request is made at the graphql_www endpoint which is used by many different features and I am not experienced enough to figure out how to hook the network layer.

Edit: The better course of action would be to inject a mark as read button in the chat long press options or to inject a small icon that does the task, however, like I mentioned, this does not stop the person on the other end from receiving the update.

swakwork:
> I understand now what you mean to say. As for what this patch is supposed to do is make the chat not appear read to the user and the person messaging them. I would think that there is no easy way to disable just the read receipts for the person messaging the user of this patch while the user is able to mark the chat as read client side.
> 
> The various mods of Instagram that I've used have injected a custom button to the chat long press menu in the inbox named "mark as read" which does exactly what it sounds like. However, unless the user pressed that, the chat would appear unread no matter the number of times the user opens to read it.
> 
> To make a client side change, the patch would be too complicated to be stable. One way we could do is to hook the method that marks the UI changes while keeping a database of the chat/thread ids that are to be marked read based on user interaction. This will have two drawbacks. 1: Like anonymously viewing stories, the change will not be persistent across sessions. 2: The patch to do this might not be feasible or if it is feasible may not be stable enough to be maintained.
> 
> One way I would've tried to implement such a patch would be to hijack the network layer and read the outgoing request body for a specific field which is `fb_api_req_friendly_name`. When it is = `IGDirectItemSeenMutation`, we can identify, by looking at the variables field which chat is being requested to be marked read. Then we maintain a custom database where we store this id field while dropping/nullifying that request. However this request is made at the graphql_www endpoint which is used by many different features and I am not experienced enough to figure out how to hook the network layer.
> 
> Edit: The better course of action would be to inject a mark as read button in the chat long press options or to inject a small icon that does the task, however, like I mentioned, this does not stop the person on the other end from receiving the update.

Please tell me you wrote it not the AI

EncryptedCicada:
> Please tell me you wrote it not the AI

I did
I was already typing this when I saw your message and just copy pasted this block from the previous message I was editing T_T

swakwork:
Screenshot_20260412-142738

The latest versions of Instagram already has options to control read receipts and typing indicator, per user bases.

EncryptedCicada:
> The latest versions of Instagram already has options to control read receipts and typing indicator, per user bases.

Yes I've seen that, but that means that the user will also not get an update when the other party reads their message, or so I think. That is what other meta apps do at least. I'll check real quick

EncryptedCicada:
Yep, just as I expected. Turning off read receipts on just one end of the chat turns it off both ways for both users so neither can see when you've read their message

yosif1705:
I agree with @EncryptedCicada that keeping the white highlight is the way to go. It’s the only real indicator for what you still need to get back to. Like you mentioned, trying to fake a "read" state locally with a database would be a headache and wouldn't persist across sessions anyway.

The "Mark as read" button is the best fix for the clutter, but it shouldn't be a local UI hack. The button should just fire the real seen request so it updates for both people. That way you stay ghosted by default but can manually sync when you're ready.

One thing tho, does replying to a chat still auto-sync the seen state on current builds? If I'm messaging back, I'm already "breaking cover," so the seen receipt should probably just fire then.

My only concern is that a global block on the mutation (here and in PR #958) might kill that future button and the reply sync too. We’d need the hook to distinguish between the automatic "chat opened" trigger and an intentional one (like a button or a reply), otherwise any manual "mark as read" feature won't actually work.

EncryptedCicada:
> One thing tho, does replying to a chat still auto-sync the seen state on current builds? If I'm messaging back, I'm already "breaking cover," so the seen receipt should probably just fire then.
> 
> My only concern is that a global block on the mutation (here and in PR #958) might kill that future button and the reply sync too. We’d need the hook to distinguish between the automatic "chat opened" trigger and an intentional one (like a button or a reply), otherwise any manual "mark as read" feature won't actually work.

Replying to a chat does "break cover" in theory but from what I've observed through testing, is that the read update is not synced when replying so we might as well force a sync then.

PR #966, opened 2026-04-13
fix(Instagram): keep one navigation button visible

### LINKED ISSUE #943: Prevent disabling all navigation buttons in Instagram
Not a bug per se but if you disable all navigation buttons like a dumdum, Instsgram will be blank. Should probably prevent disabling everything and urge to have one nav button left, especially to edit Piko settings.

You can fix it by not adding the patch, exporting Piko settings, changing the hide profile button inside the xml and then importing it before re-patching with the patch enabled

officialasishkumar:
## Summary
- prevent the Instagram navigation settings from accepting a state where every navigation button is hidden
- keep the runtime navigation filter from returning an empty button list when imported preferences are already invalid

Fixes #943

## Validation
- `git diff --check`
- Gradle build was not rerun on this branch because the same root project settings already failed before source compilation in the sibling worktree: `app.morphe.patches` could not be resolved from the Morphe GitHub Packages registry with the available token


swakwork:
The easiest solution would be removing the option to hide user profile from nav bar.

officialasishkumar:
Removed the profile-hide option from settings and stopped filtering the profile nav item at runtime, so imported prefs still keep one navigation button visible. Force-pushed the branch.

Issue #992, opened 2026-04-19
[instagram] crashes n couldn't recover


Ammar-Abdulla:
The problem started when I tried to open a photo sent that was for one view, couldn't open it so tried to save the photo and the it could tried more one more time and it crashed and couldn't be open again, just closes instantly and I got an bug report as well, tried clearing cash but to no avail https://github.com/user-attachments/assets/b54b3131-7be0-4109-a921-eda9df102477

![Insta crashed .jpg](https://github.com/user-attachments/assets/c4012666-be74-45bd-96f0-6b61dce36259)

swakwork:
Should be fixed in [v3.3.0-dev.1](https://github.com/crimera/piko/releases/tag/v3.3.0-dev.1). Do let me know

Bugdwoid:
I'm having the same issue, and it's still happening with v3.3.0-dev.1, with both Instagram versions 423 and 425.

This logcat seems to be related to the issue :

```
AndroidRuntime: FATAL EXCEPTION: main
AndroidRuntime: Process: com.instagram.android, PID: 22820
AndroidRuntime: java.lang.VerifyError: Verifier rejected class com.instagram.model.reels.ReelItem: java.lang.String com.instagram.model.reels.ReelItem.A0d(android.content.Context) failed to verify: java.lang.String com.instagram.model.reels.ReelItem.A0d(android.content.Context): [0x13] Rejecting invocation, long or double parameter at index 0 is not a pair: 4 + 1. (declaration of 'com.instagram.model.reels.ReelItem' appears in /data/app/~~1DRHTkNWyFQGOZpPSr7lbg==/com.instagram.android-wuFE1qmF07OOUtP6rD2FMQ==/base.apk)
AndroidRuntime: 	at X.4en.A0P(:828)
AndroidRuntime: 	at X.4en.A0E(:7)
AndroidRuntime: 	at X.NAC.invoke(:34)
AndroidRuntime: 	at X.KcE.A05(:72)
AndroidRuntime: 	at X.KcE.A03(:41)
AndroidRuntime: 	at X.cek.Eua(:54)
AndroidRuntime: 	at X.Pyv.invoke(:16)
AndroidRuntime: 	at X.NDK.A00(:61)
AndroidRuntime: 	at X.OQb.A02(:149)
AndroidRuntime: 	at X.OQb.A02(:112)
AndroidRuntime: 	at X.Snz.F1n(:111)
AndroidRuntime: 	at X.5dQ.F1n(:22)
[...]

Issue #969, opened 2026-04-14
Instagram keeps crashing once logged in


ghost:
`type: crash
osVersion: google/tegu/tegu:16/BP4A.260205.001/2026040801:user/release-keys
flags: dev options enabled
package: com.instagram.android:2147483647, targetSdk 36
process: com.instagram.android
processUptime: 286760 + 508 ms
installer: com.android.shell

java.lang.VerifyError: Verifier rejected class com.instagram.model.reels.ReelItem: java.lang.String com.instagram.model.reels.ReelItem.A0c(android.content.Context) failed to verify: java.lang.String com.instagram.model.reels.ReelItem.A0c(android.content.Context): [0x13] Rejecting invocation, long or double parameter at index 0 is not a pair: 4 + 1. (declaration of 'com.instagram.model.reels.ReelItem' appears in /data/app/~~o0GNS2kPp_wuhVZQwqoHoQ==/com.instagram.android-hXoeR8FOpwxGmpS7_lFC0A==/base.apk!classes4.dex)
	at X.4A4.A0P(:828)
	at X.4A4.A0E(:7)
	at X.9XA.invoke(Unknown Source:33)
	at X.1vb.A05(Unknown Source:87)
	at X.1vb.A03(Unknown Source:42)
	at X.3jD.EsY(Unknown Source:51)
	at X.9Nz.invoke(Unknown Source:216)
	at X.3jL.A00(Unknown Source:69)
	at X.A05.A02(Unknown Source:49)
	at X.A05.A02(Unknown Source:15)
	at X.5OA.Ezj(Unknown Source:119)
	at X.5Oz.Ezj(Unknown Source:22)
	at X.5le.run(:58)
	at android.os.Handler.handleCallback(Handler.java:1070)
	at android.os.Handler.dispatchMessage(Handler.java:125)
	at android.os.Looper.dispatchMessage(Looper.java:333)
	at android.os.Looper.loopOnce(Looper.java:263)
	at android.os.Looper.loop(Looper.java:367)
	at android.app.ActivityThread.main(ActivityThread.java:9331)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:566)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:837)
	at Z.init(1776158200252-3e3b1c1e-9918-4558-fd4f-909d9b8d951f)`

[...]

Issue #955, opened 2026-04-11
Instagram crashes on android 17 beta 3


davcoss:
Only works once after fresh install and can't be opened again after

dary1zhu:
Same thing happened here, I though I did something wrong on adding patches first. SInce another person has this problem on A17B3, I will provide my crash log of instagram here. Device is pixel 10 pro xl.
[crash-com-instagram-android-11_04-18-04-31_194.log](https://github.com/user-attachments/files/26645600/crash-com-instagram-android-11_04-18-04-31_194.log)

vlaarx:
I had the same on A17B3, but after disabling the follow back indicator it stopped crashing. Now on A17B4 it doesn't work at all.

swakwork:
Can you try [v3.3.0-dev.1](https://github.com/crimera/piko/releases/tag/v3.3.0-dev.1) and let me know

marianmargraf:
> Can you try [v3.3.0-dev.1](https://github.com/crimera/piko/releases/tag/v3.3.0-dev.1) and let me know

Just repatched all patches except the analytics patch. Still crashes after login on my P8P running A17 Beta4. I can enter DMs via Homescreen shortcut and it doesn't crash. But anything else force closes the app.

swakwork:
What does the log say ?

dary1zhu:
> What does the log say ?

This is after I pressed the Profile tab and after I opened app again:
AndroidRuntime: FATAL EXCEPTION: main
AndroidRuntime: Process: com.instagram.android, PID: 23880
AndroidRuntime: java.lang.VerifyError: Verifier rejected class com.instagram.model.reels.ReelItem: java.lang.String com.instagram.model.reels.ReelItem.A0d(android.content.Context) failed to verify: java.lang.String com.instagram.model.reels.ReelItem.A0d(android.content.Context): [0x13] Rejecting invocation, long or double parameter at index 0 is not a pair: 4 + 1. (declaration of 'com.instagram.model.reels.ReelItem' appears in /data/app/~~owegKKuggB6DVQJfhdiqHg==/com.instagram.android-cC5M5xfOdnR6suRjg7wCNg==/base.apk)
AndroidRuntime: 	at X.4en.A0P(:828)
AndroidRuntime: 	at X.4en.A0E(:7)
AndroidRuntime: 	at X.OdA.invoke(:34)
AndroidRuntime: 	at X.KbG.A06(:72)
[...]

Issue #947, opened 2026-04-10
GrapheneOS profile crash


Joel-dk:
I've tried multiple versions, including 423.0.0.47.66 (382907094) to no success. It'd start with a crash when clicking on profile and the app wasn't able to launch again until I cleared all data, disabling customise story timestamp patch allowed me to open the app after a "profile crash". I've tried disabling all but one patch at a time for a couple of them (disable ads etc.) and still cant click on profiles, including others peoples profiles. I've attempted all the usual GrapheneOS fixes also.

[...]

PR #1002, opened 2026-04-21
fix(instagram): correct dex emission bugs surfaced by recent ART

realies:
Three bytecode-emission bugs in Instagram patches. Recent ART Mainline
modules reject the affected classes, crashing the app the first time
the affected code path is exercised.

### Changes

1. **`StoryButton.storyButtonAction`** — declared virtual but every
   emitted callsite uses `invoke-static`. The body never references
   `this`, so adding `static` matches both the opcode and the method's
   intent.
   Throws `IncompatibleClassChangeError` at method resolution on the
   first tap of a story-action menu button.

2. **`instagramButtonEntity` emission** — issues `invoke-virtual` on an
   `Object` register where the callee expects a specific reference type.
   Inserted `check-cast p1, $buttonStyleClass` before the invoke so the
   narrowing is explicit.
   Throws `VerifyError` when `InstagramButton` is first loaded — on
   opening a user profile. dex2oat logs the rejection in advance:
   `register has type Object but expected Reference …`.

3. **`customiseStoryTimestampPatch` emission** — passes a `long (J)`
   argument as a non-consecutive register pair because the high half
   was hardcoded to `v1`. Switched to
   `invoke-static/range {v$N .. v$(N+1)}`: the pair is consecutive
   regardless of which register the low half lands in, and the `3rc`
   16-bit register encoding handles allocations where
   `postedTimestampRegister == v15` (at which point `v16` would
   overflow the non-range `35c` form's 4-bit register fields).
   Throws `VerifyError` when `ReelItem` is first loaded — on home feed
   render. dex2oat log: `long or double parameter at index 0 is not a
   pair`.

### Testing

Rebuilt the MPP from this branch, repatched Instagram on two builds,
and installed fresh on a device whose ART Mainline module enforces
these checks:

- `423.0.0.47.66` (build `382907094`) — 35/35 patches applied, 0 failed.
- `425.0.0.0.0` (build `383100026`) — 35/35 patches applied, 0 failed.

Before the fixes, dex2oat reported two `failed to verify` entries
(commits 2 and 3) and `ReelItem`/`InstagramButton` threw `VerifyError`
on feed render / profile open; tapping a story-action menu button
threw `IncompatibleClassChangeError` (commit 1). After the fixes:
dex2oat reports no rejections against our extension classes, no runtime
`VerifyError`/`IncompatibleClassChangeError`, and all three previously-
crashing flows complete — verified end-to-end on both builds.

### Closes

- #914, #969, #992 — `ReelItem.A0c` register-pair crash (commit 3).
- #880, #921, #947 — profile-tab `UI.pikoSettingsButton` VerifyError.
  Predate the v3.3.0-dev.1 `78d1b69a` refactor that consolidated the
  `IgdsButton.setStyle` injection into `InstagramButton.setStyleObject`;
  the consolidated path is fixed by commit 2, covering every current
  caller (`UI.pikoSettingsButton`,
  `ProfileMoreOption.addProfileMoreOptionsButton`).

Relates to #955 (thin report without a stack trace, likely same family).

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/story/StoryButton.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/story/StoryButton.java
@@ -55,7 +55,7 @@ public class StoryButton {
         return buttonList;
     }
 
-    public boolean storyButtonAction(CharSequence buttonText, Context ctx, Object mediaObject){
+    public static boolean storyButtonAction(CharSequence buttonText, Context ctx, Object mediaObject){
         try {
             if (buttonText.equals(Strings.VIEW_STORY_MENTIONS)) {
                 ViewStoryMentionsPatch.viewMentions(ctx, mediaObject);
--- a/patches/src/main/kotlin/app/crimera/patches/instagram/entity/instagramButton/InstagramButtonEntity.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/instagram/entity/instagramButton/InstagramButtonEntity.kt
@@ -28,10 +28,17 @@ val instagramButtonEntity =
             SetStyleExtensionFingerprint.changeFirstString(classNameToExtension(buttonStyleClass))
 
             SetStyleObjectExtensionFingerprint.method.apply {
+                // p1 arrives typed as java.lang.Object (the setStyleObject(Object)
+                // signature is intentionally generic so callers don't depend on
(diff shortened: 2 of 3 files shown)


nick490u8585:
Damn son. @swakwork look at this EFN tier commit.

realies:
oops, nuked the fork, reopening as #1004

Issue #1001, opened 2026-04-20
[Instagram] Add option to enable amoled theme


NMBlanco:
Instagram sadly removed the ability to have amoled black colors through the app (prism colors experiment)

It would be very nice if there could be a patch to enable amoled theme within Instagram app 💛

PR #1003, opened 2026-04-21
feat(instagram): add amoled theme patch

realies:
Resource patch. Remaps Instagram's dark-mode greys to pure black on
AMOLED panels by overriding four design-system color tokens.

### Changes

In `res/values-night/colors.xml`, three elevated/secondary surface tokens
are remapped to `@color/bds_black`:

- `igds_secondary_background`
- `igds_elevated_background`
- `igds_elevated_highlight_background`

`igds_primary_background` already resolves to `bds_black` upstream, so
the app background itself is unchanged. `igds_elevated_separator` is
deliberately left at `bds_grey_8` so hairline dividers remain visible
against black surfaces.

In `res/values/colors.xml`, `igds_prism_black` is overridden to
`#ff000000`. This token is IG's "Prism" design-system black, hardcoded
to `#ff0c1014` (a near-black with a slight blue tint). It has no
night-mode variant, so it renders the same in both themes — and IG's
modern feed/header/nav surfaces use it directly, which is why the
elevated overrides above don't reach them. In light mode the token is
used for text/icons where pure black is also correct.

### Testing

Rebuilt the MPP from this branch and repatched Instagram on two builds:

- `423.0.0.47.66` (build `382907094`) — 36/36 patches applied, 0 failed.
- `425.0.0.0.0` (build `383100026`) — 36/36 patches applied, 0 failed.

Verified visually in dark mode on emulator: status bar, header, story
tray, feed cards, nav, notification bands, search field, comment
threads, dialogs, DM thread list, and profile tabs all sample to
exactly `#000000`. Hairline dividers remain visible.

Android's own system permission dialogs render with the OS theme and
stay grey — out of scope for this patch.

### Closes

- #1001 — Add option to enable amoled theme

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,9 +175,6 @@ java -jar cli.jar patch --patches piko.mpp input.apkm
 <tr><td><code>Remove empty bottom space</code></td><td>Removes empty space below bottom navigation bar</td></tr>
 <tr><td><code>Disable typing status</code></td><td></td></tr>
 <tr><td><code>Hide notes tray</code></td><td>Hides notes tray in DM section</td></tr>
-<tr><td><code>Disable screenshot detection</code></td><td>Disables screenshots detection in DM</td></tr>
-<tr><td><code>Disable screenshot detection</code></td><td>Disables screenshots detection in DM</td></tr>
-<tr><td><code>Disable screenshot detection</code></td><td>Disables screenshots detection in DM</td></tr>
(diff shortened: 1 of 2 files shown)


cyriotic3:
is it possible to make a Monet version of this maybe?

realies:
It looks possible, but it's out of scope for this PR.

PR #1011, opened 2026-04-22
added comments to every feature

thyr0101:
I've only modded and dealt with instagram mods before so apologize if X's descriptions I added are off

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,17 +142,17 @@ java -jar cli.jar patch --patches piko.mpp input.apkm
 </tr>
 </thead>
 <tbody>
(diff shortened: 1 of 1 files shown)


swakwork:
Make sure you resolve the conflict. Ideally there shouldn't any conflict if you had pulled from the latest `dev`

If you don't know what handling conflict is, better you close it and create a new one by pulling the latest `dev` branch

kitadai31:
Not all patches need to have a patch description.
If the patch functionality is clear from its name, it is recommended to omit the patch description.
This is recommended by the official ReVanced Patches project.

PR #1017, opened 2026-04-23
Updated some readme descriptions

thyr0101:
Sorry for the many pull requests I haven't really done that before. Instagram & X patches' descriptions are all added (And removed a duplicate). Someone commented about not adding obvious descriptions but I believe it is more visually appealing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ java -jar cli.jar patch --patches piko.mpp input.apkm
 </tr>
 </thead>
 <tbody>
-<tr><td><code>Add ability to copy media link</code></td><td></td></tr>
+<tr><td><code>Add ability to copy media link</code></td><td>Adds the ability to copy media link.</td></tr>
 <tr><td><code>Bring back twitter</code></td><td>Bring back old twitter logo and name</td></tr>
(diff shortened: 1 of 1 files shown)


ivystopia:
The readme uses imperative mood. To be consistent, you should say "Add" not "Adds"

Since you're updating the line anyway, why not fix the British spelling "customise" to bring it in line with the other lines saying "customize"

kitadai31:
The patches list in README.md is now generated automatically.
1ffddb611c53d7e23cb0fdadd645739a35609538
Therefore, it is no longer possible to manually modify the patch list.

LisoUseInAIKyrios:
The readme patch descriptions are pulled from the [patch descriptions](https://github.com/crimera/piko/blob/4de6f3290789227eec1e85d19145d74982e8f303/patches/src/main/kotlin/app/crimera/patches/twitter/interaction/downloads/copyMediaLink/CopyMediaLink.kt#L43).

swakwork:
Sorry for the late reply, ReadMe description is now automated which makes this PR obsolete.
Thanks for the PR though :)

PR #1021, opened 2026-04-24
Better Settings 2.0

XAlboX:
Adjusted settings to get less headaches from unskilled users


swakwork:
You can literally change it in settings, I don't know why it's needs to be changed from the code

XAlboX:
This is not for me yk...
It’s for people who don’t check the settings after installing the app and instead go straight to Telegram to complain that their DMs stay unread or the reshare button is missing 💀

PR #1035, opened 2026-04-26
feat(instagram): add bytecode-level FLAG_SECURE bypass for v426

milkywaystorm:
### Description
This PR introduces a robust, surgical bytecode-level bypass for `FLAG_SECURE` on Instagram. Unlike traditional runtime layout listeners that can cause UI lag or consistency crashes, this implementation directly targets and neutralizes the internal manager responsible for enforcing window security flags.

### Technical Implementation
* **Fingerprint Targeting:** Uses the `"Inconsistency in window FLAG_SECURE state detected!"` string fingerprint to precisely locate the central `FLAG_SECURE` consistency manager class.
* **Bytecode Injection:** * Injects a Smali gate (`if-eqz v0, :original` -> `return-void`) at the entry of the consistency-check method, tied to the user's preference.
  * Scans and neutralizes all reference-counting methods within the same class (specifically, `void` methods taking `Landroid/view/Window;` as their first parameter) to prevent the flag from being dynamically reapplied.
* **UI Integration:** Fully wired into the Piko settings framework. Added a dedicated toggle under the "Ghost" section with the appropriate constants and preference getters (`Strings.java`, `Settings.java`, `Pref.java`, `SettingsStatus.java`, and `ScreenBuilder.java`).

### Why this approach?
By neutralizing the app's internal security manager directly at the bytecode level, we bypass the screenshot/screen-record protection at its root source. This prevents the framework from ever attempting to apply the flag to the window, ensuring a completely stable experience without triggering Instagram's internal state mismatch crashes.

### Testing & Validation
- **Target APK:** Instagram v426.0.0.37.68 (Build 383207302)
- **Result:** Screenshots and screen recording are flawlessly enabled in DMs and Vanish Mode when the toggle is active. No crashes, visual glitches, or instability observed.

---
### 🤖 Disclaimer: Vibe Coding
Please note that this implementation was developed using a **"Vibe Coding"** approach alongside AI (**Claude 3 Opus via Cursor**). While the logic has been thoroughly tested locally, verified via Morphe Manager, and works perfectly on the target build, I kindly ask the maintainers to review the bytecode injection logic to ensure it perfectly aligns with the project's long-term architectural standards and smali generation conventions.

### DIFF
--- a/patches/src/main/kotlin/app/crimera/patches/instagram/utils/Constants.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/instagram/utils/Constants.kt
@@ -26,7 +26,7 @@ object Constants {
                     // Stable
                     AppTarget(
                         version = "426.0.0.37.68",
-                        description = "Instagram Stable version (all archs)",
+                        description = "Tested on version code - 383207248, 383207252",
                     ),
                 ),
         )


swakwork:
If the sole purpose of this patch is blocking screenshot detection, we already have a patch for it `Disable screenshot detection` in [ScreenshotDetectionPatch.kt](https://github.com/crimera/piko/blob/dev/patches/src/main/kotlin/app/crimera/patches/instagram/misc/privacy/DisableScreenshotDetectionPatch.kt). Instead of creating 2 different patch, with 2 different toggles, doing the same stuff, combine your patches with the existing patch. Users can handle it with a single toggle

PR #1057, opened 2026-04-29
ci: Update GitHub actions, fix UK English text of 'Bring back Twitter'

LisoUseInAIKyrios:
- Updates Crowdin pull action (fixes broken cron task after this commit reaches main branch)
- Fixes UK English not showing 'Bring back Twitter' strings
- Adds missing stubbed language files expected by the Crowdin string validator


crowdin.yml
LisoUseInAIKyrios, line 9: These declarations may not be needed because the `dest` declaration below already uses Android locale format:

`translation: patches/src/main/resources/addresources/values-%android_code%/shared/strings.xml`

But it does no harm leaving them.

LisoUseInAIKyrios:
It's outside the scope of this PR, but the Crowdin project can add the remaining languages found in the files but is not on Crowdin yet:
https://github.com/crimera/piko/tree/main/patches/src/main/resources/addresources

LisoUseInAIKyrios:
You'll want to manually upload the 'Bring back Twitter' translation to Crowdin before merging this, otherwise the existing translations here on GitHub will be wiped out on the next pull from Crowdin.

kitadai31:
We do not need to upload 'Bring back Twitter' to Crowdin.
The BBT strings are now extracted directly from the strings.xml from the stock Twitter 9.98.0 APK (the last version before X), so it no longer needs to be translated by volunteers by hand.
BBT strings are already available in all languages ​​supported by X.

Also, stub files should not be necessary.
Since I added all the languages ​​for BBT, there are many untranslated languages, including those not supported by piko's Crowdin project, so I thought that adding stub files would be redundant.
Therefore, I changed the code to silently ignore them even if stub files are not present.

Please refer to #854.

Additionally, for Arabic translations, please prioritize "ar-rSA" if you combines them.
Within the stock X app, Arabic for all regions except Western Sahara (EH) uses the default "ar". (In Crowdin, an "ar" without a region code is mapped to "ar-rSA".)

LisoUseInAIKyrios:
The stub files are needed for the Crowdin string validator.

I'll make the other changes you listed.

LisoUseInAIKyrios:
Which "bring back Twitter" Arabic files are correct? Or are both needed?

There are two Arabic language folders.

Update: I see there are two different original texts for the variants. I restored the prior regional Arabic.

PR #1088, opened 2026-05-06
update(Instagram - Translations): Update Korean

BlackGold8282:


swakwork:
You might need to include in [LanguageInstance.java](https://github.com/crimera/piko/blob/3b5cf0e6777361e1fd7044bbe1c253dfd5975b00/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/translations/LanguageInstance.java#L23) as well

PR #1197, opened 2026-05-28
Add option to use Pictures folder

Vagelis1608:
Instead of Downloads.
Enabled by default.

Seems to need an app restart to apply.
Im not sure if I missed anything there.

Can also be used as a base for a proper folder picker? XD


Vagelis1608:
I just saw that its already implemented in the dev branch.

PR #845, opened 2026-03-18
refactor: replace dynamic string composition with explicit string resources

krvstek:
This PR refactors `SettingsAboutFragment` by replacing the `strEnableRes` and `strRemoveRes` methods with direct `strRes` calls. 

Previously, the code dynamically composed strings (e.g., adding `Enable...` or `Remove...` prefixes to feature names). This approach caused significant localization issues due to varying grammar rules, genders, and word orders across different languages. By using explicit string resources, the translation process becomes much easier and the UI will look cleaner across all supported languages.

PR #1200, opened 2026-05-28
refactor: Use `StringRef.str` directly

krvstek:
Follow-up to https://github.com/crimera/piko/pull/845#issuecomment-4189708560.

Adds a static import for `StringRef.str` across extension classes and replaces all `StringRef.str(...)` / `strRes(...)` wrapper call sites with the statically-imported `str(...)`.

### DIFF
--- a/extensions/twitter/src/main/java/app/morphe/extension/twitter/Utils.java
+++ b/extensions/twitter/src/main/java/app/morphe/extension/twitter/Utils.java
@@ -122,16 +122,6 @@ public class Utils {
         return value;
     }
 
-    public static String strRes(String tag) {
-        try {
-            return str(tag);
-        } catch (Exception e) {
-
-            app.morphe.extension.shared.Utils.showToastShort(tag + " not found");
-        }
-        return tag;
-    }
-
     public static void showRestartAppDialog(Context context) {
         AlertDialog.Builder dialog = new AlertDialog.Builder(context);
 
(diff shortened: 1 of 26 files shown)


extensions/twitter/src/main/java/app/morphe/extension/twitter/Utils.java
swakwork, line 1: One reason why I kept this way is to find out if any of the string value is missed out, instead of the app getting crashed, we get an error toast and the string will be displayed as the string key value

swakwork:
Lemme know if its still in Draft or ready for merger

extensions/twitter/src/main/java/app/morphe/extension/twitter/Utils.java
krvstek: I need to verify this behavior again since I haven't touched it recently. From what I remember, `ResourceUtils` is supposed to show a toast on missing strings anyway, but I'll confirm that's actually what happens.

krvstek: I temporarily removed 2 strings and checked both English and a localized version. Can confirm it's working and the app doesn't crash.

<details><summary>Screenshots</summary>
<p>

Screenshot_20260608-151335
Screenshot_20260608-151328
Screenshot_20260608-151331


</p>
</details>

krvstek:
> Lemme know if its still in Draft or ready for merger

It's ready to merge.

PR #1215, opened 2026-05-30
fix(Twitter - Customize Navigation Bar items): Sometimes the order of navigation buttons is reset

inotia00:
### Fixed an issue where the order of navigation buttons would sometimes be reset.

- There was an issue where bookmark buttons would not be displayed on the navigation bar, or the order of navigation buttons would be reset.

- In `11.88.0-release.0`, a new method for managing navigation bar items was added.

- The issue was fix by patching this method.

### Fixed an issue where the patch failed in 11.95.0-alpha.0+.

- The patch has been rewritten according to the latest Morphe Patcher syntax.

- Additionally, there were cases where the patch did not work due to [experimental flags](https://github.com/Swakshan/X-Flags/blob/a70477107e862755810448092ddcd3cb76107582/flags/x/x_flags_android_alpha.json#L4795) added in 11.95.0-alpha.0; this has also been fixed.

*Tested versions: 11.69.0-release.0, 11.88.0-release.0, 11.93.0-release.0, 11.95.1-release.0.*

### DIFF
--- a/extensions/twitter/src/main/java/app/morphe/extension/twitter/patches/FeatureSwitchPatch.java
+++ b/extensions/twitter/src/main/java/app/morphe/extension/twitter/patches/FeatureSwitchPatch.java
@@ -6,6 +6,8 @@
 
 package app.morphe.extension.twitter.patches;
 
+import static app.morphe.extension.twitter.patches.VersionCheckPatch.IS_11_95_OR_GREATER;
+
 import app.morphe.extension.twitter.Pref;
 import app.morphe.extension.twitter.Utils;
 import app.morphe.extension.twitter.settings.Settings;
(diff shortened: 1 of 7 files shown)


swakwork:
Thanks for the help. 
is it a draft or ready to be merged in the next release ?

cyberboh:
@inotia00 As i know, 11.82+ contains pairiplib which make the app crash if modified, how did you do mod 11.95.1-release.0 without any crashes?

swakwork:
> 11.82+ contains pairiplib which make the app crash if modified, how did you do mod 11.95.1-release.0 without any crashes?

At the moment we might need to rely on a ripped version  for 11.82+ but let's just say there are works going to patch without the need of a ripped version.

cyberboh:
> > 11.82+ contains pairiplib which make the app crash if modified, how did you do mod 11.95.1-release.0 without any crashes?
> 
> At the moment we might need to rely on a ripped version for 11.82+ but let's just say there are works going to patch without the need of a ripped version.

Thanks, it sounds good.
Implement pairiplib bypasser using patches is better, it saves a lot of time for lazy person like me, just patch app from apkmirror, install and forget ✨

inotia00:
> Thanks for the help. is it a draft or ready to be merged in the next release ?

It is okay to merge now.

PR #1259, opened 2026-06-05

Manishrdy:
## What

Adds a new Instagram patch — **Rename package** — that renames the patched app's
package and visible label so it can be installed alongside the original
Play Store Instagram instead of replacing it.

## Why

Currently the only way to use a patched Instagram is to uninstall the
official one. Many users would prefer to keep both — official IG for
"normal" use, patched for the customisations they want.

## How it works

- **User-facing `bytecodePatch`** carries two options: `packageName`
  (default `com.pikogram.android`) and `appLabel` (default `Pikogram`).
  It rewrites `com.instagram.*` string literals in dex but skips known
  Java class names (collected dynamically from each dex's type
  descriptors), so reflection via `Class.forName` still resolves.
- **Internal `resourcePatch`** runs in `finalize{}` and edits the
  manifest: rewrites word-boundary `com.instagram` references in
  identifier attributes (authorities, permissions, intent actions,
  launcher activity-alias labels) and overrides the string resource the
  application label already references. Class-reference attributes
  (`<application android:name>`, `<activity android:name>`, …) are left
  alone so the dex class lookup still works.
- The new namespace prefix is derived from the user's package by
  stripping the last segment, so sibling sub-namespaces
  (`com.instagram.fileprovider`, `com.instagram.barcelona.*`,
  `com.instagram.contentprovider.*`, etc.) also move under the clone —
  this is what unblocks side-by-side install with Play Store IG.
- Requires the user's package to have at least 3 segments (validation
  enforced).

## Testing

Tested on `com.instagram.android` v426.0.0.37.68 on a Galaxy S25 Ultra:
- Patching completes without smali errors.
- App launches past `WaitingForStringsActivity`, reaches the sign-in
  screen, login works.
- Manifest claims new sibling-namespace authorities under the user's
  prefix, so the official Play Store Instagram installs alongside
  without conflicts.
- Home-screen icon, install dialog, and App Info all show the user's
  chosen label.

## Known caveats

- No Play Store updates for the clone (expected for any rename).
- A few Meta cross-product integrations (FCM push from Meta's servers,
  Family Center, deeplinks targeting `com.instagram.android`) may
  degrade silently — not crashes.
- Play Protect may warn on install since the icon matches Instagram's
  but the signing cert is the patcher's.

### DIFF
--- a/patches/src/main/kotlin/app/crimera/patches/instagram/misc/renamePackage/RenamePackagePatch.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/instagram/misc/renamePackage/RenamePackagePatch.kt
@@ -1,0 +1,441 @@
+/*
+ * Copyright (C) 2026 piko <https://github.com/crimera/piko>
+ *
+ * See the included NOTICE file for GPLv3 §7(b) terms that apply to this code.
+ */
+
+package app.crimera.patches.instagram.misc.renamePackage
+
+import app.crimera.patches.instagram.utils.Constants.COMPATIBILITY_INSTAGRAM
+import app.morphe.patcher.extensions.InstructionExtensions.instructions
+import app.morphe.patcher.extensions.InstructionExtensions.replaceInstruction
+import app.morphe.patcher.patch.bytecodePatch
+import app.morphe.patcher.patch.resourcePatch
+import app.morphe.patcher.patch.stringOption
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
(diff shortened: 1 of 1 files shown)


swakwork:
1. Please create PR to dev branch 
2. I feel there exists a lot of AI slop. Please validate it once, by validate i don't mean where the app works or not, validate the code.
3. I believe cloning might need some smali code changes as well, not just package name change, check that part once.
4. Change unwanted or over explained comments into human readable comments.
5. How different is it from the change package name patch that Morphe/other Morphe-aided patches provide.
6. Do a regression testing once, especially the story part

Manishrdy:
Thanks for the review. I’ll create a PR against the dev branch.

PR #1261, opened 2026-06-05

Manishrdy:
## What this adds

A new Instagram patch — **Rename package** — that lets the user install the patched build alongside the original Play Store Instagram, by renaming the package and label at patch time.

Two patch options:
- `packageName` (default `com.pikogram.android`) — must be a valid Java package with at least 3 segments
- `appLabel` (default `Pikogram`) — display name

## How it works

The patch derives a namespace prefix from the user's package by stripping the last segment (`com.pikogram.android` → `com.pikogram`) and applies it to every word-boundary `com.instagram` reference. That covers:

- the app package itself (`com.instagram.android.*`)
- sibling sub-namespaces (`com.instagram.fileprovider`, `com.instagram.barcelona.*`, `com.instagram.contentprovider.*`, `com.instagram.quicksnap.*`, etc.) — these are system-wide-unique ContentProvider authorities; **renaming them is what unblocks side-by-side install with Play Store IG**
- custom permissions, intent actions, notification channel ids, launcher activity-alias names

Implementation splits into a user-facing `bytecodePatch` (carries the options, rewrites dex string literals) and an internal `resourcePatch` (rewrites the manifest in `finalize{}`, derives the application-label string-resource name dynamically from `<application android:label>`).

## Smali / dex behaviour (re: "is this just a package name change?")

It does rewrite at the smali level — every `const-string` literal across every dex file is checked against the namespace regex, with one important exception. A pre-pass indexes every `com.instagram.*` class type descriptor in the dex (`Lcom/instagram/X/Y;`), and the rewriter skips any string literal that exactly matches one of those dotted names. Without that skip, reflection calls like `Class.forName("com.instagram.process.instagram.InstagramApplicationForMainProcess")` (which `InstagramAppShell.onCreate` actually does) would crash at startup, because we never rename type descriptors — only literals.

What the patch does **not** add is method-body hooks (intercepting `getPackageName()`, bypassing signature checks, replacing method bodies, injecting helpers). Held off on those because adding hooks without a known failure is speculative. Happy to add targeted hooks if the maintainers see something specific that breaks.

## Manifest-side selectivity

Class-reference attributes are skipped during the manifest rewrite, since the actual Java classes stay at their original `com.instagram.*` paths in dex:

- `<application>` — `android:name`, `android:backupAgent`, `android:appComponentFactory`, `android:zygotePreloadName`
- `<activity>` — `android:name`, `android:parentActivityName`
- `<service>`, `<receiver>`, `<provider>`, `<instrumentation>` — `android:name`
- `<activity-alias>` — `android:targetActivity`
- `<meta-data>` — `android:value` (commonly a class ref for libraries like Google Cast)

`<activity-alias android:name>` intentionally stays in the rewrite — that's a logical component identifier, not a real Java class.

## Tested

`com.instagram.android` v426.0.0.37.68 on a Galaxy S25 Ultra:
- Patching completes without smali errors
- Clone launches past `WaitingForStringsActivity`, reaches sign-in, login works
- Official Play Store Instagram installs alongside without ContentProvider authority conflicts
- Story regression: viewing, posting from camera and gallery, notifications, tap-to-open, replies — all pass
- Home-screen icon, install dialog, and App Info all read the chosen label

## Known caveats

- No Play Store updates for the clone (expected for any package rename)
- A few Meta cross-product integrations (FCM push from Meta's servers, Family Center, deeplinks targeting `com.instagram.android` externally) may degrade silently — not crashes
- Play Protect may warn on install since the icon matches Instagram's but the signing cert is the patcher's

## What's left out

- Dynamic version support: the patch reads the application-label string-resource name from the manifest at patch time, so the label override works across IG versions without a hardcoded resource id. Otherwise the patch is structure-driven and should survive most IG version bumps.

### DIFF
--- a/patches/src/main/kotlin/app/crimera/patches/instagram/misc/renamePackage/RenamePackagePatch.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/instagram/misc/renamePackage/RenamePackagePatch.kt
@@ -1,0 +1,303 @@
+package app.crimera.patches.instagram.misc.renamePackage
+
+import app.crimera.patches.instagram.utils.Constants.COMPATIBILITY_INSTAGRAM
+import app.morphe.patcher.extensions.InstructionExtensions.instructions
+import app.morphe.patcher.extensions.InstructionExtensions.replaceInstruction
+import app.morphe.patcher.patch.bytecodePatch
+import app.morphe.patcher.patch.resourcePatch
+import app.morphe.patcher.patch.stringOption
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
+import com.android.tools.smali.dexlib2.iface.instruction.ReferenceInstruction
+import com.android.tools.smali.dexlib2.iface.reference.StringReference
+import org.w3c.dom.Element
+import org.w3c.dom.Node
+
+private const val ORIGINAL_PACKAGE = "com.instagram.android"
+private const val ORIGINAL_NAMESPACE_PREFIX = "com.instagram"
+private val PACKAGE_NAME_REGEX = Regex("^[a-z][a-z0-9_]*(\\.[a-z][a-z0-9_]*){2,}$")
+private val NAMESPACE_WORD_REGEX =
+    Regex("(?<![A-Za-z0-9_])" + Regex.escape(ORIGINAL_NAMESPACE_PREFIX) + "(?![A-Za-z0-9_])")
+
+private const val MAIN_ACTION = "android.intent.action.MAIN"
+private const val LAUNCHER_CATEGORY = "android.intent.category.LAUNCHER"
+
+private val SKIP_CLASS_REFERENCE_ATTRS: Set<Pair<String, String>> =
+    setOf(
+        "application" to "android:name",
+        "application" to "android:backupAgent",
+        "application" to "android:appComponentFactory",
+        "application" to "android:zygotePreloadName",
+        "activity" to "android:name",
+        "activity" to "android:parentActivityName",
+        "service" to "android:name",
+        "receiver" to "android:name",
+        "provider" to "android:name",
+        "instrumentation" to "android:name",
+        "activity-alias" to "android:targetActivity",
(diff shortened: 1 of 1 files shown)


Manishrdy:
Tested the existing Change package name patch (default morphe) with all three options enabled (Update permissions, Update providers, Update providers strings). After patching with it, the official Instagram from the Play Store still fails to install — Android rejects it because ContentProvider authorities under com.instagram.* sibling namespaces are still claimed by the patched app.

The existing patch renames the app package and its directly-declared provider names (anything under com.instagram.android.*). It doesn't touch sibling sub-namespaces like com.instagram.fileprovider, com.instagram.contentprovider.* (4 of them), com.instagram.barcelona.*, com.instagram.liteprovider.*, com.instagram.quicksnap.seenstate, com.instagram.foabackuptoken.*, com.instagram.creation.drafts.*, and com.instagram.growth.xav.*. Those are 13 ContentProvider authorities Instagram declares outside the app's own package, and Android treats every authority as system-wide-unique. As long as the patched app holds them, Play Store Instagram can't claim them.

The patch in this PR derives a namespace prefix from the user's package (strips the last segment) and applies it to every word-boundary com.instagram match across both manifest and dex. That's what specifically unblocks side-by-side install. Verified on device: official IG installs cleanly with the clone present.

Manishrdy:
Noticed a bug, app crashing while swiping across 5 or more stories at once. I'll check and open a new PR. Closing this for now.

kitadai31:
This patch contains too many unnecessary changes.

Actually, I was also investigating changing Instagram's package name.
It works with even fewer changes.

For example, regarding the strings in the dex files, it's sufficient to simply replace the exact string "com.instagram.android" with the new package name.
That's all it takes to fix video playback in Stories.

Furthermore, there's no need to edit strings.xml, nor is there any need to require more than three segments in the new package name.
Since the inter-Meta app communication functionality won't work anyway due to the different signatures, there's no need to change the prefixes of other package names like "barcelona" .

Manishrdy:
Will push a slimmer version that drops the strings.xml edit and the 3 segment requirement.

PR #1262, opened 2026-06-06

Manishrdy:
## What

Adds a Rename package patch under Instagram. The patched build installs alongside the official Play Store Instagram instead of replacing it — different package name, different signing cert, two icons on the home screen.

Two options: `packageName` (default `com.pikogram.android`) and `appLabel` (default `Pikogram`).

## How

Two patches.

A user-facing `bytecodePatch` rewrites `com.instagram.*` string literals across every dex file. A pre-pass indexes every `com.instagram.*` class type descriptor in the dex first, and the rewriter skips literals that exactly match — that keeps `Class.forName("com.instagram.X")` calls working.

An internal `resourcePatch` rewrites the manifest in `finalize{}`: package attribute, `com.instagram` references in identifier attributes (authorities, permissions, intent actions), and launcher activity-alias labels. Class-reference attributes like `<application android:name>`, `<activity android:name>`, `<activity-alias android:targetActivity>`, and `<meta-data android:value>` are left alone, because the Java classes those point at still live at their original `com.instagram.*` paths in dex.

The two patches share package + label state via top-level vars; the bytecode side writes, the resource side reads in `finalize`. Morphe's resource and bytecode contexts are separate APIs so this is the path of least resistance for a feature that needs both.

## Differences from the prior attempt (#1261)

Took the feedback and slimmed:
- Dropped the `strings.xml` application-label override. App Info reads "Instagram" again; home-screen icon still shows the chosen label via the activity-alias path. Cosmetic regression only.
- Dropped the 3-segment package-name requirement. Standard Java package format is now accepted.
- Slimmed the option-description text.

Kept what the review flagged as unneeded, but that on-device testing shows is load-bearing:
- The broad `com.instagram` namespace rewrite (not just `com.instagram.android`). Sibling sub-namespaces like `com.instagram.fileprovider`, `com.instagram.contentprovider.*`, `com.instagram.barcelona.*`, `com.instagram.liteprovider.*` and others are ContentProvider authorities. Android enforces system-wide uniqueness on authorities at install time. If the patched app keeps claiming them, installing the official Instagram alongside fails with an authority-already-claimed error. Verified on device both ways: with the narrow rewrite Play Store install fails, with the broader rewrite it succeeds.

## Smali / dex behavior

This isn't just a manifest rename. Every `const-string` literal across the dex is checked against the namespace regex and rewritten when it matches, with the class-name spare above keeping reflection refs intact. There's no method-body hook (no `getPackageName` interception, no signature-check bypass, no injected helper class). Held off on those because the cases we tested work without them. Happy to add targeted hooks if specific failures show up.

## Tested

`com.instagram.android` v426.0.0.37.68 on a Galaxy S25 Ultra:
- Patching completes
- Clone launches, login works
- Story regression flow (view, post from camera and gallery, notifications, tap-to-open, reply) — no crashes across 3 minutes of active use, confirmed via logcat
- Official Play Store Instagram installs alongside without authority conflicts
- The only errors in logcat are `SecurityException` from `com.facebook.katana.liteprovider.FirstPartyUserValuesLiteProvider` and `com.facebook.messaging.liteprovider.FamilyAppsUserValuesLiteProvider` rejecting calls from the clone. Those are FB and Messenger refusing the clone because their providers pin on Meta's signing cert. No clone can pass those checks regardless of approach.

## Caveats

- No Play Store updates for the clone (expected for any package rename)
- Meta cross-product features that pin on the signing cert won't work (FB Family Center, FB account backup, Family of Apps integrations) — visible as the SecurityExceptions above
- Play Protect may warn on install since the icon and identity match Instagram but the cert doesn't match Meta's

### DIFF
--- a/patches/src/main/kotlin/app/crimera/patches/instagram/misc/renamePackage/RenamePackagePatch.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/instagram/misc/renamePackage/RenamePackagePatch.kt
@@ -1,0 +1,258 @@
+package app.crimera.patches.instagram.misc.renamePackage
+
+import app.crimera.patches.instagram.utils.Constants.COMPATIBILITY_INSTAGRAM
+import app.morphe.patcher.extensions.InstructionExtensions.instructions
+import app.morphe.patcher.extensions.InstructionExtensions.replaceInstruction
+import app.morphe.patcher.patch.bytecodePatch
+import app.morphe.patcher.patch.resourcePatch
+import app.morphe.patcher.patch.stringOption
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
+import com.android.tools.smali.dexlib2.iface.instruction.ReferenceInstruction
+import com.android.tools.smali.dexlib2.iface.reference.StringReference
+import org.w3c.dom.Element
+import org.w3c.dom.Node
+
+private const val ORIGINAL_PACKAGE = "com.instagram.android"
+private const val ORIGINAL_NAMESPACE_PREFIX = "com.instagram"
+private val PACKAGE_NAME_REGEX = Regex("^[a-z][a-z0-9_]*(\\.[a-z][a-z0-9_]*)+$")
+private val NAMESPACE_WORD_REGEX =
+    Regex("(?<![A-Za-z0-9_])" + Regex.escape(ORIGINAL_NAMESPACE_PREFIX) + "(?![A-Za-z0-9_])")
+
+private const val MAIN_ACTION = "android.intent.action.MAIN"
+private const val LAUNCHER_CATEGORY = "android.intent.category.LAUNCHER"
+
+private val SKIP_CLASS_REFERENCE_ATTRS: Set<Pair<String, String>> =
+    setOf(
+        "application" to "android:name",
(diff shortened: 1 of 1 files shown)


kitadai31:
There are still too many unnecessary changes being made to AndroidManifest.xml.

Instead of blindly replacing "com.instagram", you should investigate the minimum necessary changes to avoid package conflicts with the official app.

Also, in smali, all that's needed is to replace the instruction "com.instagram.android", so why is the bytecode patch so complicated? What the hell is going on there?

And the source code is still AI sloppy.

Btw, I already know the minimum changes required, so I can help if you need it.
Or I might create a pull request myself. This might be a better option for both of us, as it would help avoid duplication of work.
(The reason I didn't create the patch and submit a pull request before you was simply because I was prioritizing other tasks.)
In other words, I have the advantage of already knowing how to write this patch without using AI (letting the AI ​​infer the minimum necessary changes).

Manishrdy:
Understood, closing this one. Thank you for the time on this review. Happy to test against my device if useful when your PR is up.

kitadai31:
Thank you for your understanding.

I know that contributing via pull requests is supposed to be whoever takes it first gets it.
Despite that, I'm taking over your pull request, so I'm sorry.

Now that you’ve entrusted this to me, I’ll work on it with responsibility!

PR #1265, opened 2026-06-06
perf(Instagram - Customise story ring size): Clarify standard size in the setting

kitadai31:
After the default settings value for this patch has been changed to 70, some users didn't know how to revert to the stock Instagram size, so I added the original value and instructions on how to set it in the settings screen.

Also, I removed the unnecessary ".0f" from the default value.
Since neither "." nor "f" can be entered in this setting, the ".0f" was meaningless.
I could allow decimal point input, but nobody would want to set the size down to the decimal point.

Furthermore, I changed `Float.valueOf()` to `Float.parseFloat()` because the IDE warned that the former was useless because it returns a non-primitive `Float` object.
I confirmed `parseFloat()` can parse "70.0f" and the size changes correctly.

---

p.s. This is the first time I've seen the code for the Instagram patches (Kotlin side).
I noticed that you're not using `instructionMatches` at all.
Do you know Morphe's instructionMatches feature?
This allows each element of the instruction filter of the fingerprint to be used later within a patch.

To demonstrate a practical use case for it, I've refactored this patch using instructionMatches.
Before the introduction of instruction filters, it was necessary to match a specific literal value (in this case, `100.0f`) using a custom fingerprint and then use `indexOfFirstLiteralInstruction` within the patch to search for that literal again.
But now, if you specify an instruction in the filter, you can directly access the filter's matched instructions within the patch.
Therefore, you can write the patch more concisely.
It's no longer needed to extract the literal `100.0f` as a constant.

Another example is my [ImportExportLoginTokenPatch.kt](https://github.com/crimera/piko/blob/b2ca52c692016d2808917b48be3bb9e14010e1e3/patches/src/main/kotlin/app/crimera/patches/twitter/misc/login/ImportExportLoginTokenPatch.kt#L39-L41).
Here, I used `it.instructionMatches.last().index` to add an instruction immediately after the move-result-object of the layout inflator calling and receive that result.

Instruction filters are one of Morphe's most distinctive features compared to the old Patcher.
<sup>It was also the direct cause of the split between ReVanced and Morphe.</sup>
This makes my experience writing patches in Morphe enjoyable.
I assume you will continue to develop new patches for Instagram, you can incorporate this new feature when you create a new patch.

For more details, please see here. (There is a point to note [when adding instructions for multiple matches](https://github.com/MorpheApp/morphe-patcher/blob/main/docs/2_2_1_fingerprinting.md#:~:text=Be%20careful%20if%20making%20more%20than%201%20modification%20to%20the%20same%20method%2E))
https://github.com/MorpheApp/morphe-patcher/blob/main/docs/2_2_1_fingerprinting.md

I just wanted to let you know in case you didn't know it yet.

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/translations/DefaultStrings.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/translations/DefaultStrings.java
@@ -111,6 +111,7 @@ public class DefaultStrings {
     public static String UNLOCK_PLUS_BENEFITS = "Unlock Plus benefits";
     public static String UNLOCK_PLUS_BENEFITS_DESC = "Unlock 'Plus' subscription benefits that are checked locally. USE IT AT YOUR OWN RISK";
     public static String CUSTOMISE_STORY_RING_SIZE = "Customise story ring size";
+    public static String CUSTOMISE_STORY_RING_SIZE_DESC = "Enter the percentage of story ring size. 100 is the standard and maximum size.";
 
     public static String CATEGORY_DOWNLOAD_MEDIA = "Download media";
     public static String ENABLE_DOWNLOAD = "Enable download";
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/translations/DefaultStrings.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/translations/DefaultStrings.java
@@ -111,7 +111,7 @@ public class DefaultStrings {
     public static String UNLOCK_PLUS_BENEFITS = "Unlock Plus benefits";
     public static String UNLOCK_PLUS_BENEFITS_DESC = "Unlock 'Plus' subscription benefits that are checked locally. USE IT AT YOUR OWN RISK";
     public static String CUSTOMISE_STORY_RING_SIZE = "Customise story ring size";
-    public static String CUSTOMISE_STORY_RING_SIZE_DESC = "Enter the percentage of story ring size. 100 is the standard size.";
+    public static String CUSTOMISE_STORY_RING_SIZE_DESC = "Change the size of story ring by a percentage (Default is 100)";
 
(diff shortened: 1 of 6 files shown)


swakwork:
Ya I never tried `instructionMatches` in the past, so sticked to basics. Will learn about it and start using it in the future ones 👍

extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/translations/DefaultStrings.java
```diff
@@ -111,6 +111,7 @@ public class DefaultStrings {
     public static String UNLOCK_PLUS_BENEFITS = "Unlock Plus benefits";
     public static String UNLOCK_PLUS_BENEFITS_DESC = "Unlock 'Plus' subscription benefits that are checked locally. USE IT AT YOUR OWN RISK";
     public static String CUSTOMISE_STORY_RING_SIZE = "Customise story ring size";
+    public static String CUSTOMISE_STORY_RING_SIZE_DESC = "Enter the percentage of story ring size. 100 is the standard and maximum size.";
```
swakwork: 100 is the default one but isn't the maximum size per say. I tried increasing it to 200 and it did work and the app didn't crash.
You can probably change it to just standard size

kitadai31: Really? Values above 100 do not work on my end. Whether it's 120 or 200, the size is the same as 100.

However, if it works on your end, this might be an A/B test or something. I can remove "maximum" from the description.

swakwork: Based on DPI or something ig. On my tab, it reflects aka increases to 200+ but on my phone 100 is the max.

kitadai31:
Please don't merge for now
I came up with a better strings
will update tomorrow

swakwork:
Ready ?

PR #1288, opened 2026-06-10
feat(Instagram): Add Spanish translation

Oleepre:
Hopefully this pull request is done correctly. Everything was translated and double-checked to make sure it sounds fine and in line with terms used in other patches, like the original Morphe's ones, for this language.


swakwork:
Could you please update the copyright tag to match the updated one. 

https://github.com/crimera/piko/blob/b5f64fe4a93fe2c199116615aa5e509d6c89cbad/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/translations/DefaultStrings.java#L1-L5

Oleepre:
Everything should be up-to-date now.

Issue #1037, opened 2026-04-26
[Instagram] Stories video does not play when using Change Package Name patch on Samsung devices


TaubeKiste-ai:
### Description

First of all, the other patches generally works fine.

The issue only appears in a specific setup on Samsung devices. Since Instagram cannot be fully uninstalled on many Samsung devices, the patched app cannot be installed normally with the original package name. Because of that, the `Change Package Name` patch has to be used.

I tested this on the following devices:

- Samsung Galaxy S22 Ultra
- Samsung Galaxy S24 Ultra

When the patched app is installed using the `Change Package Name` patch, Instagram mostly works, but Stories are broken: story videos do not play visually, only the audio plays.

This seems related to the issue described here:
https://github.com/crimera/piko/issues/879

### Steps to reproduce

1. Use a Samsung device where the original Instagram app cannot be uninstalled.
2. Patch Instagram with Piko.
3. Enable/use the `Change Package Name` patch.
4. Install the patched APK.
5. Open Instagram and play a Story containing video.

### Expected behavior

Story videos should play normally, including both video and audio.

### Actual behavior

Only the audio plays. The video itself does not play/show.

### Notes

Without the Samsung package-name conflict, the patch generally works as expected. The problem seems to be triggered when the `Change Package Name` patch is required.

joaovoliv:
I have an S23 and have the same problem. Tried everything possible and couldn't make it work. Looking forward to an update that will fix this issue.

PS: I tried messing arround with some cache options in developer setting. Don't ever bother, it just made me loose my time

Issue #1164, opened 2026-05-21
make it a separate app from instagram


Kurizek27:
so i can use piko alongside regular instagram with newest features

leyyce:
You can already do that. Just activate the "Change package name" patch.

kvxrun:
> You can already do that. Just activate the "Change package name" patch.

Doing that introduces another bug where videos on stories don't play

leyyce:
> > You can already do that. Just activate the "Change package name" patch.
> 
> Doing that introduces another bug where videos on stories don't play

Well if that's the case that's unfortunate and hopefully someone can introduce a fix for it, but changing the package name literally is the only way to make Android recognize two applications as being different. It's the defacto way for Android to differentiate between apps. If two apps have the same package name they'll always be treated as the same app and they will basically override each other (if they're signed with the same key. Otherwise it will try to act like an app update but fail because of the different signatures). Conversely, if they have different package names they'll be treated as different apps you can alongside each other, even if they're the exact same app otherwise.

As far as I know there is no other way to install two apps side by side because that's just how Android is built. For Android same package name equals same app, thus it will go into "update mode" instead of "install mode" if an app with that package name already exists.

Kurizek27:
> You can already do that. Just activate the "Change package name" patch.

![Screenshot_2026-09-15-17-23-12-369_app.morphe.manager.jpg](https://github.com/user-attachments/assets/156ae9f4-6edf-4062-9698-e4ce8f4fc64e)

Where exactly??

leyyce:
> > You can already do that. Just activate the "Change package name" patch.
> 
> 
> Where exactly??
> 

You need to enable expert mode in the settings first

![Settings](https://imglink.cc/cdn/Mn5OfcwFCp.jpg)


And then you can select the _Clone_ patch. Make sure you're selecting the one in the _Piko_ tab and not the one under _Morphe Patches_

![Patch Screen](https://imglink.cc/cdn/7Nh3eqtFlf.jpg)

Issue #1098, opened 2026-05-08
feat(Instagram): Add patch to Change package name


imPratyoosh:
### Feature description

A patch which changes patched APK package name to be installed along side with original unpatched Instagram

### Why this patch matters? 

Few reasons are listed here

+ To be able to install & use both patched and unpatched app at once (not everyone wants but could be useful for some)

+ Can have multiple sets of patched apps, for debugging purposes & analyzing flags etc

+ HyperOS(tested) & other OS(haven't tested) limit fps/refresh rate of some known apps when videos play, Instagram is one of them, this cap lowers the refresh rate of the app ui to 30fps, which feels very bad while navigating through reels
This patch would definitely solve this problem as I've tested it with a Foss Instagram mod **Instafel** which offers both clone and unclone apk, performance on clone version is significantly better

`Device running on Android 13 HyperOS 1`

kvxrun:
You can currently do this via using other patch sources along with piko, but that breaks videos on stories.

SoansSandha:
> You can currently do this via using other patch sources along with piko, but that breaks videos on stories.

what other patch source? iv also tried this but no luck, here is [my post](https://github.com/crimera/piko/issues/1094) with how i tried it

kvxrun:
> what other patch source? iv also tried this but no luck, here is [my post](https://github.com/crimera/piko/issues/1094) with how i tried it

devanced/morphe universal patches work for me, but then again, videos on stories freeze up

SoansSandha:
> devanced/morphe universal patches work for me, but then again, videos on stories freeze up

just tried and i cant even install the cloned app -_-

Issue #1094, opened 2026-05-07
(Help) How can i patch insta as a clone?


SoansSandha:
I have been trying to install insta with the patches as a clone. So i have all the Piko patches, and also from the Morphe Patches iv selected "change package name" (worked for multiple other apps, for me to clone). but when installing it says "Package conflict" "you need to uninstall the existing version"

in the "change package name" patch, iv also try to enable "Update permissions", "update providers" and "update provider strings". but still no luck

nick490u8585:
package name change should be enough.

imPratyoosh:
@SoansSandha you can track feature request [here](https://github.com/crimera/piko/issues/1098)

imPratyoosh:
@SoansSandha please close this as completed

PR #1295, opened 2026-06-11
feat(Instagram): Add `Clone` patch

kitadai31:
Supersedes PRs: #1259 #1261 #1262

Closes: #1037 #1094 #1098 #1164

This patch changes the package name and app name (label) of the app.
The differences from the official Morphe's "Change package name" universal patch are:

- Properly rename permissions and providers to prevent conflicts with the official app.
- Fixes the issue where video stories only play audio.
- Contains custom branding name functionality.

### About the name "Clone"

Initially, I was planning to follow the ReVanced/Morphe convention and create two patches: "Change package name" and "Custom branding name".
However, for the following reasons, I decided to combine them into a single “Clone” patch.

- In the Instagram modding community, the term "Clone" is commonly used.
- My guess is that no one would change the package name without changing the app name. So, there's no point in making the two patches selectable independently.

### Other matters to be discussed

Currently, the default values for this patch are as follows:

- Package name: `com.instagram.android.piko`
- App name: `Piko Instagram`

However, @Manishrdy had used the following name in a previous pull request:

- Package name: `com.pikogram.android`
- App name: `Pikogram`

In my case, since the name "Pikogram" is not used anywhere in the official piko project, I used the simpler name "Piko Instagram".
However, if Swak likes the name "Pikogram," we can switch to that instead, or any other names.
Please note that "Instagram Piko" is not recommended, as the AOSP launcher is likely to omit the string and make it indistinguishable as "Instagra...".

Also, this patch can includes the custom branding app icon feature in the future. But since Piko Instagram does not currently have any branding (logo), this is outside the scope of this pull request.

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/UI.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/UI.java
@@ -18,8 +18,8 @@ import android.view.ViewGroup;
 import android.widget.ImageView;
 import java.util.ArrayList;
 
+import app.morphe.extension.crimera.PikoUtils;
 import app.morphe.extension.instagram.utils.Pref;
-import app.morphe.extension.instagram.settings.Settings;
 import app.morphe.extension.instagram.entity.InstagramDialogBox;
 import app.morphe.extension.instagram.settings.ActivityHook;
 import app.morphe.extension.shared.Logger;
@@ -169,7 +169,7 @@ public class UI {
                         String selectedOption = options.get(which);
 
                         if (selectedOption.equals(Strings.GOTO_PIKO_SETTINGS)) {
-                            ActivityHook.openLink("instagram://profile");
+                            PikoUtils.openUrl("instagram://profile");
                         }
                     }
                 } catch (Exception e) {
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/MediaData.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/MediaData.java
@@ -85,7 +85,7 @@ public class MediaData extends Entity {
         return new UserData(userData);
     }
 
-    public HashSet getMentionSet() throws Exception {
+    public HashSet<UserData> getMentionSet() throws Exception {
         Object result =  super.getMethod(this.getExtendedData(), "methodName");
         if (result != null) {
             List userInteractionList = (List) result;
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/story/ViewStoryMentionsPatch.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/story/ViewStoryMentionsPatch.java
@@ -8,22 +8,16 @@
 package app.morphe.extension.instagram.patches.story;
 
 
-import java.lang.reflect.Field;
-import java.util.List;
-import java.util.Set;
 import java.util.HashSet;
 import android.app.Dialog;
 import android.content.Context;
-import android.content.DialogInterface;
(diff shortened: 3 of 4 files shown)


kitadai31:
@Manishrdy Hi, there are two things I want to apologize to you for.

First, I told you that you shouldn't need to edit strings.xml, but I was wrong. (https://github.com/crimera/piko/pull/1261#issuecomment-4636999302)
When I was researching changing the package name, I hadn't yet considered changing the app name. Therefore, I wasn't aware that Instagram sets labels for many activities.
Ultimately, I decided to edit strings.xml because replacing the strings within the xml would be more reliable.

Second, I told you that you only need to replace the string "com.instagram.android" in the bytecode. (https://github.com/crimera/piko/pull/1262#issuecomment-4638162802)
This is because I actually tested the app and didn't find any bugs.
However, just to be safe, I decided to replace the provider name as well.

My patch should be safer because it only replaces provider names that are actually found. However, if you know of any other strings that need to be replaced to prevent a functionality from breaking, please let me know.

swakwork:
Thanks for the PR. is it WIP or ready to merge ?

if its ready to merge I was wondering can it be expanded to X/Twitter as well ? or can it be worked upon in the later future

kitadai31:
I've just found one issue and fixed
Ready to merge

> I was wondering can it be expanded to X/Twitter as well ?

Idk, but it is likely more difficult than Instagram. Twitter app relies heavily on content providers internally, makes it difficult.

> can it be worked upon in the later future

This patch is expected to work in the future.
There are no version-dependent elements.
If there is any change, it would be a change in how to fix video stories.

swakwork:
> > can it be worked upon in the later future
> 
> This patch is expected to work in the future. There are no version-dependent elements. If there is any change, it would be a change in how to fix video stories.

Was talking about expanding it to Twitter in the future. But ya got your answer here

> Idk, but it is likely more difficult than Instagram. Twitter app relies heavily on content providers internally, makes it difficult.


Lemme merge this one then 🤝 thanks

swakwork:
One last question, Just curious, you think we may/can update the app icon as well ? Ik we don't have a specific theme/logo for `Piko`

kitadai31:
> you think we may/can update the app icon as well ?

Yes. The only problem preventing this is that piko doesn't have a theme color or logo yet.
We can replace icons, similar to the Bring back Twitter patch or the RVX's Custom branding icon patch.

Manishrdy:
@kitadai31  Thanks for the detailed explanation and for taking the time to work through this.

I wanted to suggest something, given the exceptional work you and the team have done here, and the trust you've earned from the community through open source transparency, I think Pikogram / Pikoinsta is the way to go. 

Look at what works in this space, likes of, Honinsta, Instaeclipse, Instafel, Instaprime, InstaGold. They're all recognizable because they own their identity. Piko has the momentum and the community trust to carry a strong name.

kitadai31:
Excuse me for repeatedly pushing minor changes even though I had initially said that it was ready.
This time it's truly ready now on my side.

I'll leave decisions regarding branding (name and icon) entirely up to the project members.
"Pikogram", "Pikoinsta", "InstaPiko", ... Hmm, it's really hard to choose 🤔

Manishrdy:
Built and tested the latest Clone patch locally. Everything worked flawlessly on my end. Nice work, and thanks for the thorough implementation and for sharing the insights behind the decisions and tradeoffs.

ashes-xda:
I've tried patching piko patches ONLY (morphe's disabled) but after patching the apk it conflict with the original app and asks me to uninstall it first.
Am i missing something here ? 
Version 3.9.0 (with pre-release enabled)

PR #1304, opened 2026-06-12
chore(Instagram): migrate strings from `Strings.java` to XML resources

krvstek:
https://github.com/crimera/piko/pull/1327


swakwork:
Thank a bunch 🤝

swakwork:
What's the logic behind `IgStr class` ? Why is there need of a cache map ?
Also what's the logic behind the `baseContext()` ? I remember @kitadai31 talking about the same in the telegram group - [Chat link](https://t.me/pikopatches/81386/103416). If what I understood from his message is correct, he told me to use the particular class's context to get the strings.

krvstek:
> What's the logic behind `IgStr class` ? Why is there need of a cache map ? 

Calling `getIdentifier()` is notoriously slow. If we fetch strings dynamically without caching (especially in the feed or settings), it will cause noticeable UI lag and frame drops. The map just ensures we only pay that performance cost once.

> Also what's the logic behind the `baseContext()` ?

Instagram heavily wraps its UI contexts and messes with the `AssetManager`. If we use the standard Activity/View context, we run into severe Resource ID collisions, just like if we use `StringRef.str()` directly.

During my testing, fetching a simple default string from our XML via a regular UI context returned random obfuscated Meta classes (like `X.07tN`) instead of the actual string, crashing our reflection patches. Unwrapping to `baseContext()` bypasses Meta's polluted overrides and safely hits the clean Android resource table where our injected XMLs actually live.

krvstek:
Just a heads-up: I found a bug during testing. Toast messages for Download actions (like `Download profile picture`) are returning random native Instagram strings (e.g. `Category` or `Account activated`) instead of our injected XML texts. Interestingly, the `Copy` actions work perfectly fine.

This looks like a Resource ID collision where our injected IDs are overlapping with Meta's native ones. Using `baseContext()` doesn't solve it. I'm reporting this here because I'm not sure how to fix this issue. Let me know what you think.

krvstek:
Closing this, since better approach is coming soon.

Edit: nevermind.

PR #1315, opened 2026-06-14
Updated the Portuguese translation and added the Spanish translation 

ViictorNunes:
Edit: I didn't end up noticing it and you've already added the Spanish translation, so ignore the translation file.


swakwork:
> Edit: I didn't end up noticing it and you've already added the Spanish translation, so ignore the translation file.

I can't remove/ignore files while merging, so if you feel that the file is redundant, please remove it and commit. I will be squashing the commits, so don't worry about number of commit messages

Issue #739, opened 2025-11-26
Localization support whitout PR


speedlorenzo12:
I've noticed that a lot of PR is being released for translations.
Couldn't we use sites like CrowdIn for example, or a separate site like Anddea did for the patches translation?

krvstek:
This issue can be closed btw.

speedlorenzo12:
I had completely forgotten about its existence

PR #1327, opened 2026-06-15
feat(Instagram): Add Crowdin support

krvstek:
Migrates the application's text localization from hardcoded Java classes (`Strings.java` and related language-specific instances) to standard Android XML string resources in order to support translating Instagram via Crowdin.

*Tested on Instagram version 430.0.0.53.80 [383611190] (both original and localized XML strings were tested).*


swakwork:
Get the default strings like download folder name, etc that I have hardcoded be present in the `Strings.java`, 

https://github.com/krvstek/piko/blob/cbf8976471f7fa286a77dc57b74d726bdd890cc7/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/Strings.java#L17-L25

just remove those that are added to xml

krvstek:
> Get the default strings like download folder name, etc that I have hardcoded be present in the `Strings.java`,

Done.

patches/src/main/kotlin/app/crimera/patches/instagram/misc/settings/SettingsPatch.kt
```diff
@@ -85,11 +88,6 @@ val settingsPatch =
                     firstInvokeSuperIndex + 2,
                     LOAD_FLAGS_DESCRIPTOR.format("load"),
                 )
-                // Loads strings for common extension.
-                addInstruction(
-                    firstInvokeSuperIndex + 3,
-                    "invoke-static {}, $CONSTANTS_DESCRIPTOR/Strings;->load()V",
```
swakwork, line 91: You may need to restore this with the updated Constants class

krvstek:
Nothing TODO left, LGTM. It's ready for review.

extensions/instagram/src/main/java/app/morphe/extension/instagram/utils/IgStr.java
```diff
@@ -0,0 +1,57 @@
...
+        if (template.equals(name)) return name;
+        try {
+            return String.format(template, args);
+        } catch (Exception e) {
+            return template;
+        }
+    }
+
+    private static Context baseContext() {
+        Context ctx = Utils.getActivity();
```
kitadai31: It should be fine to store the retrieved base context in a static field instead of getting it every time.

Furthermore, you don't need to try `Utils.getActivity()`.
The Instagram's SharedExtensionPatch only hooks into the Application onCreate, so getActivity() will always be null.

FYI, I said "use activity context" on Telegram, but this only applies to the Piko's SettingsActivity.
There is no point in using an Activity context anywhere else.

kitadai31 (commented):
I'm planning to submit another review tomorrow, so could you please wait for the merge until then?

extensions/instagram/src/main/java/app/morphe/extension/instagram/utils/IgStr.java
swakwork: Just an FYI I have added a custom setActivity during Fragment oncreate

https://github.com/crimera/piko/blob/dee8831cc36786a5fee9d751f3270238f82830c6/patches/src/main/kotlin/app/crimera/patches/instagram/misc/settings/SettingsPatch.kt#L59

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/ScreenBuilder.java
```diff
@@ -74,76 +74,76 @@ public void buildAdsSection() {
...
                             Settings.DIRECTLY_OPEN_METACONFIG
                     )
             );
             addPreference(category,
                     helper.buttonPreference(
-                            Strings.EXPORT_DEV_OVERRIDES,
+                            str("piko_export_dev_overrides"),
                             "",
-                            Strings.EXPORT_DEV_OVERRIDES
+                            str("piko_export_dev_overrides")
```
kitadai31: The third argument of helper.buttonPreference is simply used as a key, so instead of using `str()` to localize it, I think it's better to use the string name as a key directly like `"piko_export_dev_overrides"`.
The same applies to `ButtonPref.java`.

I understand that it will work even if it's localized, but this makes the intent of the code clearer.

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/fragments/RestorePrefActivity.java
```diff
@@ -40,17 +42,17 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
...
-            if (args.containsKey(Strings.IMPORT_DEV_OVERRIDES)) {
+            if (args.containsKey(str("piko_import_dev_overrides"))) {
                 destinationFile =new File(context.getFilesDir()+ "/mobileconfig","mc_overrides.json");
-            } else if (args.containsKey(Strings.IMPORT_ID_MAPPING)) {
+            } else if (args.containsKey(str("piko_import_id_mapping"))) {
                 destinationFile = new File(context.getFilesDir()+ "/mobileconfig","id_name_mapping.json");
-            } else if (args.containsKey(Strings.IMPORT_PIKO_PREF)) {
-                destinationFile =  new File(context.getApplicationInfo().dataDir + "/shared_prefs",Strings.PIKO_SETTINGS+".xml");
+            } else if (args.containsKey(str("piko_import_pref"))) {
+                destinationFile =  new File(context.getApplicationInfo().dataDir + "/shared_prefs",Constants.PIKO_SETTINGS+".xml"); // CHANGED: fix typo (estinationFile → destinationFile)
```
kitadai31: Leftover comment?

patches/src/main/resources/addresources/values/instagram/strings.xml
```diff
@@ -0,0 +1,211 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="piko_title_settings">Piko Settings</string>
+
+    <!-- Ads Section -->
```
kitadai31, line 3: Comments may not be necessary.

The comments in Twitter's strings.xml file were a remnant from before they migrated to Crowdin.
Comments were added because they made the XML easier to read when translating it directly.

On the other hand, when translating strings on Crowdin, it makes no difference whether there are comments or not.
In Crowdin, these comments are only displayed as contextual information for a single string, so comments explaining the meaning of an entire section are not very meaningful.

patches/src/main/resources/addresources/values/instagram/strings.xml
```diff
@@ -0,0 +1,211 @@
...
+    <string name="piko_category_ads">Ads</string>
+    <string name="piko_disable_ads">Disable ads</string>
+    <string name="piko_hide_suggested_content">Hide suggested content</string>
+    <string name="piko_hide_suggested_content_desc">Hides suggested stories, reels, threads (Suggested posts will still be shown).</string>
+
+    <!-- Developer Options Section -->
```
kitadai31, line 9: I think this is a good opportunity to rearrange the order of the sections to match the actual order on the settings screen. (Just a suggestion)

The section order can be:
Ads → Ghost → Links → ……

I believe this makes translation works on Crowdin more intuitive.

kitadai31:
Also, we have to deal with `CheckString.kt`

If you don't add "instagram" here, the string check won't be performed.
However, simply adding it will result in an error. Try running `main` function of it.

To address this, you need to create an stab "instagram/strings.xml" file in the folder for each language like [LisoUseInAiKyrios did](https://github.com/crimera/piko/pull/1057#issuecomment-4343780226).

However, in my opinion, this is ridiculous.
It's ridiculous to add stub XML files for all languages ​​when far fewer languages ​​are enabled in piko's Crowdin.
I also think it unnecessarily increases the size of the patch bundle file.
Therefore, I support the option of simply changing the following line to just skip it instead of adding stubs.

```
if (stream == null) throw IllegalArgumentException("Could not find resource $srcSubPath"
```

patches/src/main/resources/addresources/values/instagram/strings.xml
krvstek: I added these comments to keep the file structure consistent with the rest of the project (e.g. the Twitter strings uses the exact same approach for string categorization).

patches/src/main/resources/addresources/values/instagram/strings.xml
krvstek: I strictly followed the order of the original English strings from `DefaultStrings.java` to maintain a 1:1 mapping and make it easier to track.

We could definitely bring the Twitter strings up to the same standard, but refactoring that is out of scope for this specific PR.

krvstek:
> If you don't add "instagram" here, the string check won't be performed. However, simply adding it will result in an error. Try running `main` function of it.
> 
> To address this, you need to create an stab "instagram/strings.xml" file in the folder for each language like [LisoUseInAiKyrios did](https://github.com/crimera/piko/pull/1057#issuecomment-4343780226).
> 
> However, in my opinion, this is ridiculous. It's ridiculous to add stub XML files for all languages ​​when far fewer languages ​​are enabled in piko's Crowdin. I also think it unnecessarily increases the size of the patch bundle file. Therefore, I support the option of simply changing the following line to just skip it instead of adding stubs.

Instead of modifying the script to just skip the check, we could follow the approach Liso took. We can allow the translators to port the existing translations out of the `.java` files (I can help with this process if needed), and then release a bundle in the future that includes these proper files.

I know adding empty stubs might seem a bit redundant right now, and yes, the patch bundle size will slightly increase, but honestly, the size difference is absolutely negligible. It's simply worth it for a cleaner, long-term translation pipeline.

patches/src/main/resources/addresources/values/instagram/strings.xml
kitadai31: Twitter comments can also be deleted in the future.

I know the story behind how comments were added to Twitter's strings.xml file.

Originally, Twitter's strings.xml file had no comments, and each section was separated by blank lines.
The content of that section was clear from the first string of the section, so a comment was unnecessary.

However, in the Japanese strings.xml file, one translator persistently deleted blank lines.
I tried to restore the blank lines to match the original strings.xml several times, but he still kept deleted them.
It appears he was uploading strings to his private Crowdin project and translating them there. After each translation, he would export the XML file from Crowdin, upload the entire file to GitHub, and submit a pull request. This is why blank lines were being lost.

So, only the Japanese strings.xml file lacked blank lines separating each category, making it difficult to read.
Eventually, another Japanese translator added XML comments because it was difficult to read. This was unnecessary if blank lines had been present. This action was wrong.

Therefore, the comments for sections should never have existed in the first place.
Now that translation is done on Crowdin, it's completely unnecessary.
Please remove it before adding any technical debt.

patches/src/main/resources/addresources/values/instagram/strings.xml
kitadai31: That makes sense 👍
At this stage, keeping the order as is makes it easier to check for any omissions during the migration from Java to XML.

kitadai31:
Was the cache really useless?

How about caching only the resource IDs?
This approach achieves performance equivalent to pre-compiled resource IDs while reducing heap memory usage compared to storing the entire string in a map.
This was my initial thought.

To be honest, I also had doubts about the effectiveness of Morphe's `StringRef` class.
I don't really understand the point of caching an object that references a string instead of the resource ID or the String itself lol
This originates from the OG Vanced, though

krvstek:
> Was the cache really useless?

The cache was a leftover from the old JSON approach. Caching `baseContext` is sufficient, and I didn't notice any measurable difference in RAM or CPU usage after dropping it.

> How about caching only the resource IDs?
> This approach achieves performance equivalent to pre-compiled resource IDs while reducing heap memory usage compared to storing the entire string in a map.
> This was my initial thought.

Is it really necessary? As far as I know, Android's own `getIdentifier()` cache handles the common case already. We'd need to profile hot paths to justify the added complexity.

kitadai31:
Oh, well then, that might not be a problem.
I was concerned because every time I use `getIdentifier()`, Android Studio keep warning me that resource reflection degrades performance 😅

And thx for removing XML comments, it's much clear!

krvstek:
> Oh, well then, that might not be a problem.
> I was concerned because every time I use `getIdentifier()`, Android Studio keep warning me that resource reflection degrades performance 😅

Even though I didn't feel any lag or stuttering during my own testing, it's probably better to play it safe, especially for older devices or heavy scrolling scenarios. I'll add a lightweight ID cache to avoid any potential performance hits in the future. Better safe than sorry, I'll push it shortly!

kitadai31 (approved):

krvstek:
By the way, I think we can link this PR to this issue: https://github.com/crimera/piko/issues/739

patches/src/main/resources/addresources/values/instagram/strings.xml
krvstek: I restored the comments. I get the history, but Swak pointed out they still need them to navigate and group new strings on the dev branch: https://github.com/crimera/piko/pull/1357#issuecomment-4733451303

swakwork:
Just wondering, do we need to add anything manually from the Crowdin or will it (Instagram folder) gets pushed automatically to Crowdin ?

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/feed/ChangeLikeAnimationPatch.java
```diff
@@ -20,10 +19,12 @@ public class ChangeLikeAnimationPatch {
...
     public static Object changeLikeAnimation(Object defaultAnimation){
         try {
-            if(CHANGE_LIKE_ANIMATION.equals(Strings.DEFAULT)) return null;
+            if (CHANGE_LIKE_ANIMATION == null || "ARES_LIKE_ACTIVATION".equals(CHANGE_LIKE_ANIMATION)) {
+                return null;
+            }
 
             Entity entity = new Entity();
-            Class<?> animationEnumClass = Class.forName("X.05zO");
+            Class<?> animationEnumClass = Class.forName("X.07tN");
```
swakwork, line 27: PS: This get modified from during patching, we got fingerprint/patch for it,  But no worries no need to revert back

krvstek:
> Just wondering, do we need to add anything manually from the Crowdin or will it (Instagram folder) gets pushed automatically to Crowdin ?

Everything you add to Instagram XML file will be automatically pushed to Crowdin.

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/feed/ChangeLikeAnimationPatch.java
krvstek: I actually changed it because it was spamming me with error toasts from `ChangeLikeAnimationPatch` during testing.

swakwork:
Could you check on the conflicts as well. If it's all set, I'm planning to merge this as the first thing, so that we can work on other updates without any conflicts happening. And also fortunately we don't have any translations PR in the queue

krvstek:
> Could you check on the conflicts as well. If it's all set, I'm planning to merge this as the first thing, so that we can work on other updates without any conflicts happening. And also fortunately we don't have any translations PR in the queue

It's ready.

PR #1330, opened 2026-06-15
Update translation portuguese and correction to make post time work.

ViictorNunes:


swakwork:
like I said previously please create separate PR for each of the update.
If you are planning on creating as a single PR please write the commit message in the following pattern so that it gets included in the changelogs as well as other contributes and understand your fix
https://github.com/crimera/piko/blob/b2ca52c692016d2808917b48be3bb9e14010e1e3/.releaserc#L26-L29

pattern:
`<type>(<app name>): <commit message>`

like:
image

ViictorNunes:
> like I said previously please create separate PR for each of the update. If you are planning on creating as a single PR please write the commit message in the following pattern so that it gets included in the changelogs as well as other contributes and understand your fix
> 
> https://github.com/crimera/piko/blob/b2ca52c692016d2808917b48be3bb9e14010e1e3/.releaserc#L26-L29
> 
> pattern: `<type>(<app name>): <commit message>`
> 
> like: image


Oh, I ended up not noticing that I was doing comitts automatically, I'll fix it when I get home.

swakwork:
Also for now dont update the flags or add new features, their stable at the moment. Lets work on those after the next stable release

ViictorNunes:
ok

PR #1333, opened 2026-06-15
fix(Instagram): Warn when external downloader package is unset

naoverse06:
When "Download with external downloader" is enabled but the external downloader package name is empty, Piko still tries to launch an `ACTION_SEND` intent with an empty package name. That results in a generic Activity launch failure message instead of telling the user what needs to be configured.
This change checks the external downloader package name before generating the share link or launching the activity. If the package name is not set or only contains whitespace, Piko now shows a toast asking the user to set the external downloader package name and stops there. If a package name is set, the existing external downloader flow continues with the trimmed package name.

## Testing
- `.\gradlew.bat :patches:checkStringResources`


swakwork:
https://github.com/crimera/piko/blob/e1d185a651a579bfa8543996df2831ca9bee97fe/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/ActivityHook.java#L25-L33

launchActivity functions already has the flags added so you can revert the changes you have made to `ActivityHook.java` file

PR #1357, opened 2026-06-17
chore: Cleanup comments in twitter strings

krvstek:
Follow-up to: https://github.com/crimera/piko/pull/1327#discussion_r3422529412

Since translations are handled by Crowdin, these section comments are no longer necessary and have been removed for better maintainability.

### DIFF
--- a/patches/src/main/resources/addresources/values/twitter/strings.xml
+++ b/patches/src/main/resources/addresources/values/twitter/strings.xml
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources>
-    <!-- Premium Settings -->
     <string name="piko_title_premium">Premium</string>
     <string name="piko_pref_undo_posts">Enable undo posts</string>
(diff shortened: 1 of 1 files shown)


swakwork:
I do rather have these if it's not causing any issues.

It's easy to add comments based on different sections. And moreover mostly new English strings are added while a patch is being added and are pushed to the dev branch along with the patches.

krvstek:
> I do rather have these if it's not causing any issues.
> 
> It's easy to add comments based on different sections. And moreover mostly new English strings are added while a patch is being added and are pushed to the dev branch along with the patches.

I actually removed them here just to keep things consistent with another PR where I also deleted section comments. If you prefer keeping them for better navigation during development, that makes sense!

Should I restore the comments in that other PR as well so we stay aligned across the board?

swakwork:
> Should I restore the comments in that other PR as well so we stay aligned across the board?

If you can, yes please

PR #1367, opened 2026-06-18
bump: Add `11.99.0-release.1` to supported versions list (Requires Piko-Shim)

kitadai31:
A very quick hotfix.

The upload of version 12.0.0 on APKMirror is delayed.
And APKs of 12.0.0 uploaded to other websites are unstable because they often lack dpi, architecture, and language splits.

The target version `11.99.0-release.1` is also included in Piko-Shim patches, so it works well when they are combined.

Tested on my phone.

### DIFF
--- a/patches/src/main/kotlin/app/crimera/patches/twitter/utils/Constants.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/twitter/utils/Constants.kt
@@ -13,25 +13,19 @@ object Constants {
             appIconColor = 0x000000,
             targets =
                 listOf(
-                    // Beta
-                    AppTarget(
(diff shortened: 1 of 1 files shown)


swakwork:
I feel like we can also remove the first 3 stable beta Alpha versions, what do you say ?

kitadai31:
I think it's fine to keep 11.81.0-release.0 for the time being.
11.69 still exists, so 11.81 can still be there.

Beta and alpha can be removed.

swakwork:
> Beta and alpha can be removed.

Cool, can you remove those in this PR itself

swakwork:
Thanks, we Good to merge?

kitadai31:
Yes

PR #1377, opened 2026-06-19
fix(Instagram): Show downloader package error on launch failure

naoverse06:
The previous external downloader warning fix handles the unset or blank package name case before launching the activity. in the final merged version of that PR, invalid package names can still show the generic "ActivityHook: launchActivity failure" message instead of the external downloader package name error
This happens because the launch failure is handled inside `launchActivity()` before `openLink()` can show its fallback toast
Adjust the external downloader launch failure handling so `openLink()` can show the proper package name error for invalid package names

## Testing
- `.\gradlew.bat :patches:checkStringResources`
- Tested on Instagram v430.0.0.53.80 with Piko v3.6.0


swakwork:
The PR doesn't add any value if you ask me.

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/ActivityHook.java
```diff
@@ -82,11 +86,12 @@ public static void openLink(String url, String packageName) {
         intent.setType("text/plain");
         intent.putExtra(Intent.EXTRA_TEXT, url);
         intent.setPackage(packageName);
-         try {
-            launchActivity(intent);
+
+        try {
+            startActivity(PikoUtils.getContext(), intent);
         } catch (Exception e) {
             PikoUtils.toast(Strings.EXTERNAL_DOWNLOADER_PACKAGE_NAME_NOT_FOUND);
```
swakwork: I don't think we can have this toast here. `openLink()` is a generic method which is not specific to external download.

naoverse06: This change is meant to cover the invalid configured package name case. i agree the toast should not live in the generic helper, so i moved it to externalDownloader() and made openLink() only report whether launching succeeded

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/ActivityHook.java
```diff
@@ -21,11 +21,15 @@
 @SuppressWarnings("deprecation")
 public class ActivityHook {
 
+    private static void startActivity(Context context, Intent intent){
```
swakwork: No need for this extra method, let it stay inside `launchActivity()`

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/ActivityHook.java
```diff
@@ -77,16 +81,18 @@ public static void openLink(String url) {
...
     }
 
-    public static void openLink(String url, String packageName) {
+    public static boolean openLink(String url, String packageName) {
         Intent intent = new Intent(Intent.ACTION_SEND);
         intent.setType("text/plain");
         intent.putExtra(Intent.EXTRA_TEXT, url);
         intent.setPackage(packageName);
-         try {
-            launchActivity(intent);
```
swakwork, line 87: remove the try catch and directly call `launchActivity()`. The exception is already handled at `launchActivity()`, having it here makes no sense.

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/download/DownloadUtils.java
```diff
@@ -246,7 +246,9 @@ public static void externalDownloader(Object mediaObject, int currentMediaIndex)
                 return;
             }
             String link = Links.generatePostLink(mediaObject, currentMediaIndex);
-            ActivityHook.openLink(link, packageName);
+            if(!ActivityHook.openLink(link, packageName)){
```
swakwork: This is not the ideal way. We may probably need to think of a different. Probably check if the user has this an app with this package name installed and then throw this toast if its not present

Like this for example
https://github.com/MorpheApp/morphe-patches/blob/d6a88edcfba71f9b630314c4c8b56347a10c8b2a/extensions/youtube/src/main/java/app/morphe/extension/youtube/settings/preference/ExternalDownloaderPreference.java#L128

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/download/DownloadUtils.java
```diff
@@ -237,6 +238,14 @@ public static void downloadMediaUrl(Context context, String mediaUrl, String sub
         downloader.enqueue(new DownloadRequest(mediaUrl, subFolder, fileName));
     }
 
+    private static boolean isAppInstalledAndEnabled(String packageName){
```
swakwork: You can probably move this to `PikoUtils.java` as it's not specific to DownloadUtils per say. We can use it other needs as well if needed in the future.


And also don't forget to add credits to Morphe. Add the permalink to the function were you have taken from.

PR #1381, opened 2026-06-20
refactor: Use switch-case for language selection

Mocha313:
Replace the long if-else chain with a switch on Locale.getDefault().getLanguage() for clarity and maintainability. Each language now maps to its corresponding Strings class, and a default case returns StringsEnglish as the fallback; removed the redundant trailing return.


swakwork:
No longer required as Locale is now handled via Crowdin since https://github.com/crimera/piko/pull/1327

Issue #832, opened 2026-03-08
Post source label on private tweets


qnqelle:
The app crashes/freezes when opening a tweet from a private account with the "show post source label" toggle enabled

Issue #641, opened 2025-08-13
App crashes when opening tweets from private accounts


DjMon33:
Tweets from private accounts I follow are displayed in the timeline, but the app consistently crashes when attempting to open them. This occurs in the current version (v.11.11.2-release.0), but also happened in previous versions I used.

kitadai31:
@Swakshan said that this is a known issue of "Show post source"

Turn off "Show post source" until it is fixed

PR #1410, opened 2026-06-24
fix(twitter) remove error on bad response code when show source label is enabled

ItsJiggles:
added a check to prevent the app from freezing/crashing when viewing a private tweet with show tweet source label patch enabled


ItsJiggles:
fixes #832

ItsJiggles:
actually i think just not erroring no matter the response code is the better solution bc i think this can also happen if you click on a deleted tweet and in general i dont think the app should crash when the request fails

ItsJiggles:
fixes #641 as well

swakwork:
Sorry for the I prioritised version update over feature updates. 

Thanks for the PR, could you check if it's possible to release the lock in 5 secs? irrespective of whether the response is recieved or not.

ItsJiggles:
sorry for the delayed response @swakwork 

at the moment, it seems like the reason it freezes up is because if fetchStatusById() ever returns null, it is indistinguishable from it never having responded. the easiest fix that i can see is just returning an empty JSONObject. there could be a more elegant solution, but this works fine I think.

PR #1500, opened 2026-07-11
refactor(Instagram): Using SAF for media download, not All Files Access

bikram-agarwal:
- Updated FolderPickerActivity to utilize the new URI-based storage access.
- Enhanced MediaDownloader to support document-based downloads.
- Improved StorageUtils to manage custom download URIs more effectively.
- Changed string resources for download path selection and permission prompts.
- Removed the obsolete FilePathHelper class.


bikram-agarwal:
Built the .mpp from this and patched recommended Instagram with that. Set a folder as download destination, it asks SAF "Allow" permission to that one folder, once granted - all media downloads there. No "All Files Access" required.

swakwork:
Please create the PR for `dev` branch. Thanks

bikram-agarwal:
> Please create the PR for `dev` branch. Thanks

Switched to dev branch.

swakwork:
Could you also check if its possible to extend support for saving in SD card as well ?

bikram-agarwal:
> Could you also check if its possible to extend support for saving in SD card as well ?

This should already be supported by this change as long as the SD card/folder is exposed by Android's system folder picker.

The new flow uses `ACTION_OPEN_DOCUMENT_TREE`, persists the selected tree URI with `takePersistableUriPermission(...)`, and writes through `DocumentsContract.createDocument(...)` + `ContentResolver.openOutputStream(...)`. It no longer resolves or writes to raw filesystem paths, so it is not tied to internal storage paths like `/storage/emulated/0`.

The only limitation is Android/provider behavior: if the system picker exposes a writable SD-card folder and grants the URI permission, downloads should work there. If a device/OEM blocks selecting a particular SD root/folder, that is outside the patch's control. I have verified the SAF flow works, but I do not have a physical SD-card device to explicitly test that provider path.

swakwork:
Oh awesome, thanks, will test and include on the next release

Issue #1046, opened 2026-04-27
[INSTAGRAM] Monet Theme


xchwarze:
There's a feature in MyInstaApp that I really liked: the Monet theme.
Could it be ported to Piko?

P.S.: It might be a good idea to use GitHub issue templates; they're useful for projects like this.

swakwork:
https://github.com/crimera/piko/pull/1509

Issue #884, opened 2026-03-30
Feature (Instagram) Material you dynamic theme


alok1nanda:
Hi. Can you please add patch for dynamic theme for Instagram

bikram-agarwal:
+1. And I hope the color can be the app's background color (like YT does), instead of just accent colors of different elements.

marcoshimselfxd:
which version (link) did you download for instagram? to run with piko
i cant make it work https://github.com/crimera/piko/issues/885

swakwork:
https://github.com/crimera/piko/pull/1509

PR #1509, opened 2026-07-12
feat(instagram): Material You theme

bikram-agarwal:
## ✨ New Features
- **Material You theme for Instagram.** New `Theme` patch option (`theme = material`, the default) that recolors Instagram from the device's Android 12+ dynamic, wallpaper-derived palette. It **follows the device's light/dark setting**; devices below Android 12 get fixed-hex fallbacks approximating the same look. Dark mode uses the neutral tonal palette (calm, low-chroma dark surfaces); light mode uses a light accent tint so the wallpaper colour is visible without washing content out.

## 🛠 Improved Features
- **Unified `Theme` patch replaces the standalone AMOLED patch.** `AmoledThemePatch.kt` is removed and `ThemePatch.kt` exposes a `theme` string option — `material` (Material You) or `amoled` (pure-black), default `material`. AMOLED behavior is unchanged, just selected through the option, which now carries a title and a bulleted description listing accepted values and the known limitations.
- **The piko settings screen follows the selected theme.** `InstagramPreferenceStyle` color helpers resolve the `piko_dynamic_*` palette at runtime (Context-aware, with dark/light fallbacks when a resource is absent). The custom preference switch is redrawn with dynamic track/thumb colors, an on-state checkmark, and a properly visible disabled state; callers in `SettingsActivity`/`CategoryPref` thread `Context` through.
- **Native preference dialogs follow the device light/dark.** List / multi-select / edit-text preference dialogs previously inherited the settings activity's fixed dark platform theme, so they rendered dark on a light device. They now build from a device-appropriate `ContextThemeWrapper` (`InstagramPreferenceStyle.dialogContext`).
- **`MultiSelectListPref` is now styled like the other prefs** (adds `onCreateView`/`onBindView`), so its rows use the themed title/summary colours instead of the near-invisible framework default.

## 📦 Others
- Removed `AmoledThemePatch.kt` (superseded by `ThemePatch.kt`).

## Implementation notes for reviewers
- **Colour-mapping model.** `ThemePatch` upserts `piko_dynamic_*` alias colors into `values`, `values-night`, `values-v31`, `values-night-v31` (dynamic `@android:color/system_*` in the v31 buckets, fixed hex otherwise), then remaps Instagram's tokens onto those aliases. Two token classes are handled deliberately differently:
  - *Literal* greyscale tokens (`bds_black`/`bds_white`, `bds_grey_*`, `igds_prism_gray_*`) → **fixed lightness constants** (identical in every bucket). Instagram reuses a literal for opposite roles per mode (e.g. `bds_black` is a dark *surface* in dark mode but dark *text* in light mode), so a constant keeps contrast correct both ways while IG's own theme picks the light-vs-dark literal per mode.
  - *Semantic* role tokens (`igds_primary_background`/`_text`/`_icon`, separators) and *accent* tokens → **mode-flipping** aliases (light values in the day buckets, dark in the night buckets).
- **Surface coverage.** Beyond IGDS, the GM3/"baseline" substrate and the MDC theme-attribute leaves are remapped for both modes — dark (`gm3_baseline_surface_dark`, `design_dark_default_color_background`, `cardview_dark_background`, `material_grey_850/900`, …) and light (`gm3_baseline_surface_light`, `abc_decor_view_status_guard_light`, `background_floating_material_light`, `material_grey_50/100`, …). The light MDC `colorSurface`/`colorBackgroundFloating` leaves were the key to theming light-mode bottom sheets, dialogs and the feed.
- **Native dialog theming.** `dialogContext()` wraps the preference context in `Theme.DeviceDefault[.Light]` so framework `AlertDialog`s follow the device. Row views are coloured explicitly, so the wrapper only affects the dialog.
- **Known limitations (documented in the option description).** The theme covers resource-backed in-app surfaces only. A few server-driven pages (notifications, DM inbox, About-this-account) and full-screen media/Reels keep Instagram's own colors in both modes — they're composited outside the resource-themeable view tree and aren't reachable by a static resource patch.

## Screenshots
MU

## Testing
- Built and installed on an Android 12+ device. Verified `material` in both light and dark device modes, and `amoled`, across the feed, bottom sheets, dialogs, menus, toasts, toolbars, buttons, and the piko settings screen (switch enabled/disabled states; native list, multi-select and edit-text dialogs).


extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
```diff
@@ -50,24 +49,75 @@ public static int dp(Context context, float value) {
         );
     }
 
-    public static int backgroundColor() {
-        return UI.getThemedColour("igds_color_primary_background");
+    public static boolean isDark(Context context) {
```
swakwork: This was present earlier but it breaks the theming of Piko Setting.
Try changing dark mode , light mode and system default. 

Thats why fixed it by using instagram's existing colors

https://github.com/crimera/piko/commit/f1f278a87665be59218dcc54af983b411802fe0f

swakwork:
Harding coding colors isnt a good idea, and also isDark was broken that why I used the colors from Instagram themself.
https://github.com/crimera/piko/commit/f1f278a87665be59218dcc54af983b411802fe0f

I do believe it would be a wiser option to change the colors in the Xml styles rather than smali

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
```diff
@@ -50,24 +49,75 @@ public static int dp(Context context, float value) {
...
-    public static int secondaryTextColor() {
-        return UI.getThemedColour("igds_color_secondary_text");
+    public static int pressedBackgroundColor(Context context) {
+        return colorResource(context, "piko_dynamic_pressed_background", isDark(context) ? Color.rgb(35, 38, 43) : Color.rgb(239, 239, 240));
     }
 
-    public static int disabledTextColor() {
-        return UI.getThemedColour("igds_color_separator");
+    public static int primaryTextColor(Context context) {
+        return colorResource(context, "piko_dynamic_on_surface", isDark(context) ? Color.rgb(245, 245, 245) : Color.rgb(9, 12, 16));
```
swakwork: No need for isDark function. Instead pass the existing themedColor as fallback

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
```diff
@@ -50,24 +49,75 @@ public static int dp(Context context, float value) {
...
+    public static int outlineColor(Context context) {
+        return colorResource(context, "piko_dynamic_outline", isDark(context) ? Color.rgb(145, 150, 162) : Color.rgb(108, 114, 123));
+    }
+
+    public static int outlineVariantColor(Context context) {
+        return colorResource(context, "piko_dynamic_outline_variant", isDark(context) ? Color.rgb(95, 99, 110) : Color.rgb(220, 222, 230));
+    }
+
+    private static int colorResource(Context context, String name, int fallback) {
+        int id = context.getResources().getIdentifier(name, "color", context.getPackageName());
```
swakwork: No need to pass Context, there is an inbuilt function to retrieve colors by passing just text in the morphe's Utils class, utilise that instead

vippium:
It will be much better if there is a selection for dark theme. Like Material + Amoled or Material + Dark. 

So that both blend perfectly and user can experience both at same time.

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
bikram-agarwal: Piko

Piko settings page (and all sub-pages) look fine in both device dark and light mode.

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
bikram-agarwal: Thanks - agreed on all three. I'll switch the settings screen back to Instagram's themed colors (UI.getThemedColour) and drop the isDark/Context/hardcoded-fallback layer I'd reintroduced; since the patch remaps the underlying color leaves, the settings pick up the dynamic colors through IG's own attributes, so that Java isn't needed. 

Three small UX bits I'd added on top are separate from the color feedback - happy to keep them, move them to a follow-up, or drop them, your call: 

1. A redrawn switch with a checkmark (I'll source its colors from IG's accent attrs, not hardcoded); 
2. A fix so the native list/edit-text dialogs follow the device light/dark instead of always-dark (it selects a platform dialog theme by device mode - no hardcoded colors); 
3. Styling the multi-select rows like the other prefs. 
 
Should I keep them in this PR, or should I move them to a separate PR?

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
swakwork: you are have patched with material so it wont be going to the fallback, which would cause the issue

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
swakwork: Those are fine I suppose, just do this one https://github.com/crimera/piko/pull/1509#discussion_r3568035090

And make sure to check all the modes

bikram-agarwal: Made the changes, and verified on my device (Pixel 9 Pro XL, running Android 17). In both device theme = dark or light, Instagram color switches correctly and app looks fine. 

Nasa

Limitations mentioned in the OP (and also in the patch description, see OP screenshot)

- A few server-driven pages (notifications, DM inbox, About-this-account) and Reels tab keeps Instagram's own colors in both modes.

And a new thing I discovered - 
On the feed page, after a reel finishes playing, the "watch more reels" button that appears at the center - that gets dark color with dark text, so you can't read what it is saying. But I think that's not much of a loss. (Also, this happens only in device dark mode.)

Watch More

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
```diff
@@ -50,6 +50,23 @@ public static int dp(Context context, float value) {
...
+     * Native list / multi-select / edit-text preference dialogs build from
+     * getContext()'s platform alertDialogTheme. The settings activity is registered
+     * with a fixed dark DeviceDefault theme, so those dialogs render dark even on a
+     * light device. Wrapping the preference's context in a device-appropriate
+     * DeviceDefault theme makes the native dialogs follow the device's light/dark
+     * setting. NOTE: this only selects a platform DIALOG theme by device mode — it
+     * does not set any app colour (those still come from Instagram's themed attrs).
+     */
+    public static Context dialogContext(Context context) {
+        int nightMode = context.getResources().getConfiguration().uiMode & Configuration.UI_MODE_NIGHT_MASK;
```
swakwork: I have doubts on this. Last time when this check was made, it failed. 
Could you check if the lists are properly colored on all the different themes? w &w/o material theme

krvstek:
#884 and #1046 should be closed, if merged.

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
bikram-agarwal: Pushed another commit for some more tweak. This is the different combination result (4 scenarios in columns) 

|  | Both Dark | Both Light | Conflicting | Conflicting |
| --- | --- | --- | --- | --- |
| **System theme** | Dark | Light | Dark | Light |
| **Insta theme** | Dark On / system default | Dark off / system default | Dark off | Dark On |
| **App pages** | Dark | Light | Light | Dark |
| **App dialogs** | Dark | Light | Light | Dark |
| **Piko dialogs** | Dark | Light | Light (but everything visible) | Light (but everything visible) |
| **Nav buttons** | Dark | Dark (Correct) | ⚠️ White (somewhat visible) ¹ | ⛔ Dark (almost invisible) ¹ |
| **Watch more reels button** | ⛔ Dark pill, dark text, so can't read | Visible | Visible | ⛔ Dark pill, dark text, so can't read |

¹ - It gets fixed if you go to reels tab (which is always dark theme in all modes) and come back. 

## Theme behavior — summary & known limitations

**Core result:** app pages and app dialogs follow **Instagram's in-app theme** (the "Insta theme" row), not the system setting - so Material You themes correctly in all four combos. With IG on **"System default"** (the default), IG always follows the device, so you're always in a "Both" column and everything is correct. The two "Conflicting" columns only occur if IG's in-app theme is deliberately set *opposite* the device.

**Piko dialogs** are themed via a platform DeviceDefault theme (following the device), so they stay fully readable in every combo — they may not perfectly match IG in the conflicting cases, but nothing is hidden.

**Two known limitations, each inherent to a static resource patch:**

- **⚠️ Nav buttons (conflicting combos only — transient, self-healing):** on first render in a deliberately-conflicting device/IG setting, the tab-bar icons can be mis-tinted (light-on-light / dark-on-dark). It **corrects itself as soon as you switch tabs** — e.g. open Reels (always dark) and come back and they render correctly. So it's a stale-initial-paint / theme-refresh timing quirk, not a permanent mis-color: the icons *do* resolve to IG's theme once refreshed. Minor and self-recovering; a real fix would need a runtime nudge to re-tint on theme resolution, which a resource patch can't do.

- **⛔ "Watch more reels" pill (whenever the app is dark):** dark pill + dark text. It's an on-media color (`igds_color_*_on_media`) that shares a single color leaf with a real app surface - one leaf can't be both a dark surface *and* light-on-media text, and the surface has to win. 

Both artifacts appear only in deliberately-conflicting device/IG settings or on media surfaces. They can be left as documented limitations.

swakwork: in human langauge pls

bikram-agarwal: In plain terms:

**Normally it just works.** As long as your phone and Instagram are set to match — both light, both dark, or Instagram on "System" (which just follows the phone) — the theme applies correctly across pages and dialogs.

**A few screens keep Instagram's own colors** and can't be themed this way — they'll look like stock Instagram:
- Notifications (activity) page
- DM inbox (the message list)
- Reels tab (it's full-screen video, so it stays dark)
- "About this account" page

**Two small quirks, _only if you set phone and Instagram to *opposite* themes_** (phone dark but IG light, or the reverse):
1. **Bottom tab icons** briefly show the wrong shade but fix themselves the moment you switch tabs (tap Reels and come back). A momentary first-load glitch, not a real coloring problem.
2. **The "Watch more reels" button** shows dark text on a dark pill when theme is dark, so it's hard to read.

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemePatch.kt
```diff
@@ -0,0 +1,489 @@
...
+@Suppress("unused")
+val themePatch =
+    resourcePatch(
+        name = "Theme",
+        description = "Applies one Instagram theme at patch time.",
+        default = true,
+    ) {
+        compatibleWith(COMPATIBILITY_INSTAGRAM)
+
+        val theme by stringOption(
```
swakwork: Can you change these to booleanOptions instead.

swakwork:
The seems to be issue with Text color across the app. Find which one color attr it is and set it to white or a lighter color

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemePatch.kt
bikram-agarwal: done

bikram-agarwal:
I see the on-media text was dark. I've now repointed the on-media attributes to white instead, leaving the surface leaf alone.

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemePatch.kt
```diff
@@ -0,0 +1,508 @@
...
+import org.w3c.dom.Element
+import java.io.FileWriter
+import java.nio.file.Files
+
+@Suppress("unused")
+val themePatch =
+    resourcePatch(
+        name = "Theme",
+        description = "Applies one Instagram theme at patch time.",
+        default = true,
```
swakwork: Set the default as false, Themeing is generally left as user preferences rather than a must have patch

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
```diff
@@ -50,6 +50,24 @@ public static int dp(Context context, float value) {
...
+     * getContext()'s platform alertDialogTheme. The settings activity is registered
+     * with a fixed dark DeviceDefault theme, so without this the dialogs are always
+     * dark. Wrapping the preference's context in a device-appropriate DeviceDefault
+     * theme makes them follow the device's light/dark setting — which matches
+     * Instagram whenever IG's in-app theme follows the device (the "System" default)
+     * or is set to match it. This only selects a platform DIALOG theme by device
+     * night mode; it sets no app colour.
+     */
+    public static Context dialogContext(Context context) {
+        int nightMode = context.getResources().getConfiguration().uiMode & Configuration.UI_MODE_NIGHT_MASK;
```
swakwork: You can now use the isDark() function to check dark mode. Utilise it

bikram-agarwal: Done.

kitadai31:
e888776cd426a6e8170e744375843b400abd86b4

The default value of `default` is `true`
You need to set `default = false` explicitly

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemePatch.kt
```diff
@@ -0,0 +1,509 @@
...
+                "follows your device's light/dark setting. Turn this on to override it " +
+                "with a pure-black AMOLED theme instead. " +
+                "Note: a few server-driven pages (notifications, DM inbox, " +
+                "About-this-account) and full-screen media/Reels keep Instagram's own " +
+                "colours in both modes.",
+        )
+
+        execute {
+            forceWhiteOnMediaChrome()
+            if (amoled == true) applyAmoledTheme() else applyMaterialYouTheme()
```
swakwork, line 42: Dont use else, create a new boolean option for Material You Theme

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemePatch.kt
```diff
@@ -0,0 +1,509 @@
...
+import org.w3c.dom.Document
+import org.w3c.dom.Element
+import java.io.FileWriter
+import java.nio.file.Files
+
+@Suppress("unused")
+val themePatch =
+    resourcePatch(
+        name = "Theme",
+        description = "Applies one Instagram theme at patch time.",
```
swakwork: You can probably change the description to something like "Applies Amoled or Material you theme, at patch time"

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemePatch.kt
bikram-agarwal: You mean let the user enable both material you an amoled at the same time?? That doesn't sound right. Turning one on should turn the other off, imo. And morphe patcher doesn't allow "turning X on should turn Y off".

swakwork: Oh wait I didnt notice you have added a description. NVM

Issue #1073, opened 2026-05-03
[Feature Request] Notify of deleted/unsent messages on Instagram


wefalltomorrow:
This was a great feature of MyInsta, would be nice to see it added to Piko patches :)

PR #1531, opened 2026-07-15
feat(instagram): save deleted messages

vishnuprasad0:
Saves incoming DMs as they arrive so you can still read them after they're unsent.
Adds a button on the DM header to view a chat's deleted messages (or all of them). Works for text and media — photos, videos, shared posts/reels and voice notes. Tapping a saved message lets you open, download or copy it.
Toggle it under DM settings.

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/DirectItem.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/DirectItem.java
@@ -1,0 +1,293 @@
+/*
+ * Copyright (C) 2026 piko <https://github.com/crimera/piko>
+ *
+ * See the included NOTICE file for GPLv3 §7(b) terms that apply to this code.
+ */
+
+
+package app.morphe.extension.instagram.entity;
+
+/**
+ * Wrapper around a single Instagram DM (DirectItem).
+ *
+ * <p>All placeholder names below (e.g. {@code "fieldName"}) are rewritten at patch time by
+ * {@code directItemEntity}, so at runtime this class only does plain reflection on names
+ * already correct for the installed app version. See {@code MediaData} for the reference pattern.
+ *
+ * <p>Reads go through the base class because the MQTT subclass shadows several base fields with
+ * same-named fields of a different type. Reading off the (patched) base class avoids that mix-up.
+ */
+public class DirectItem extends Entity {
+    private final Object obj;
+
+    public DirectItem(Object obj) {
+        super(obj);
+        this.obj = obj;
+    }
+
+    /** Binary name of the DirectItem base class (patched, e.g. {@code X.9ZA}). */
+    private String getBaseClassName() {
+        return "className";
+    }
+
+    /** Loads the base class via the app classloader. */
+    private Class<?> baseClass() throws Exception {
+        return Class.forName(this.getBaseClassName());
+    }
+
+    /** Reads a base-class field, bypassing MQTT subclass field shadowing. */
+    private Object readBaseField(String fieldName) throws Exception {
+        return super.getField(this.baseClass(), this.obj, fieldName);
+    }
+
+    public String getItemId() {
+        try {
+            Object v = this.readBaseField("fieldName");
+            return v instanceof String ? (String) v : null;
+        } catch (Exception e) {
+            return null;
+        }
+    }
+
+    public String getUserId() {
+        try {
+            Object v = this.readBaseField("fieldName");
+            return v instanceof String ? (String) v : null;
+        } catch (Exception e) {
+            return null;
+        }
+    }
+
+    public String getText() {
+        // REST stores text on the base class; MQTT leaves it null and uses a subclass payload field.
+        try {
+            Object v = this.readBaseField("baseTextField");
+            if (v instanceof CharSequence) return v.toString();
+        } catch (Exception ignored) {
+        }
+        try {
+            Object v = super.getField(this.obj, "subTextField");
+            if (v instanceof CharSequence) return v.toString();
+        } catch (Exception ignored) {
+        }
+        return null;
+    }
+
+    /** Raw server timestamp string (microseconds). */
+    private String getTimestampRaw() {
+        try {
+            Object v = this.readBaseField("fieldName");
+            return v instanceof String ? (String) v : null;
+        } catch (Exception e) {
+            return null;
+        }
+    }
+
+    /** Timestamp in milliseconds, or {@code now} when unavailable. */
+    public long getTimestampMs() {
+        String raw = this.getTimestampRaw();
+        if (raw != null && !raw.isEmpty()) {
+            try {
+                return Long.parseLong(raw) / 1000L; // microseconds -> milliseconds
+            } catch (NumberFormatException ignored) {
+            }
+        }
+        return System.currentTimeMillis();
+    }
+
+    /** True when the sender has unsent (hidden) this message. */
+    public boolean isHideInThread() {
+        try {
+            Object v = this.readBaseField("fieldName");
+            return v instanceof Boolean && (Boolean) v;
+        } catch (Exception e) {
+            return false;
+        }
+    }
+
+    /** True when sent by the logged-in user. */
+    public boolean isSentByViewer() {
+        try {
+            Object v = this.readBaseField("fieldName");
+            return v instanceof Boolean && (Boolean) v;
+        } catch (Exception e) {
+            return false;
+        }
+    }
+
+    /** Clears/sets the hide_in_thread flag (anti-revoke). */
+    public void setHideInThread(boolean hidden) {
+        try {
+            java.lang.reflect.Field f = this.baseClass().getDeclaredField("fieldName");
+            f.setAccessible(true);
+            f.set(this.obj, hidden);
+        } catch (Exception ignored) {
+        }
+    }
+
+    /** Restores the text on both the base and MQTT subclass fields (anti-revoke). */
+    public void setText(String text) {
+        try {
+            java.lang.reflect.Field f = this.baseClass().getDeclaredField("baseTextField");
+            f.setAccessible(true);
+            f.set(this.obj, text);
+        } catch (Exception ignored) {
+        }
+        try {
+            java.lang.reflect.Field f = this.obj.getClass().getDeclaredField("subTextField");
+            f.setAccessible(true);
+            f.set(this.obj, text);
+        } catch (Exception ignored) {
+        }
+    }
+
+    /** item_type as a stable token (enum {@code toString()}), or null. */
+    public String getItemType() {
+        try {
+            Object v = this.readBaseField("fieldName");
+            return v != null ? v.toString() : null;
+        } catch (Exception e) {
+            return null;
+        }
+    }
+
+    /** The DirectThreadKey object, or null. */
+    private Object getThreadKey() throws Exception {
+        return this.readBaseField("fieldName");
+    }
+
+    /** The numeric thread id this message belongs to, or null. */
+    public String getThreadId() {
+        try {
+            Object key = this.getThreadKey();
+            if (key == null) return null;
+            Object v = super.getField(key, "fieldName");
+            return v instanceof String ? (String) v : null;
+        } catch (Exception e) {
+            return null;
+        }
+    }
+
+    // Media URL resolution. Every supported shape carries an unobfuscated Media object (the stable
+    // anchor); field names below are placeholders rewritten at patch time by directItemEntity.
+    //   * DIRECT  — a Media field (media, media_share, raven_media).
+    //   * WRAPPED — a reshare/voice wrapper holding a Media (clip, reel_share, voice_media).
+    // Not resolved (no stable type anchor): animated_media, story_share, xma, link -> "[type]" label.
+
+    /** Binary name of the concrete item class (e.g. X/6fW) that declares the media fields. */
+    private String mediaClassName() { return "mediaClassName"; }
+
+    private Class<?> mediaClass() throws Exception { return Class.forName(this.mediaClassName()); }
+
+    // Field-name providers — one placeholder each, rewritten at patch time via changeFirstString.
+    private String fieldMedia()        { return "fieldName"; } // media        -> Media
+    private String fieldMediaShare()   { return "fieldName"; } // media_share  -> Media
+    private String fieldRavenMedia()   { return "fieldName"; } // raven_media  -> Media
+    private String fieldClip()         { return "fieldName"; } // clip         -> wrapper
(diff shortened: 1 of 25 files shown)


swakwork:
Thanks for the new PR. is it WIP or ready to ship?

vishnuprasad0:
Not WIP,  it's ready — can push fixes later if anything comes up. It closes #1073

extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/DirectItem.java
```diff
@@ -0,0 +1,293 @@
...
+     * Runtime scan, not patch-time resolution: an xma reshare has no Media object, so the only
+     * patch-time anchor would be opcode/iput sequencing that needs re-RE every version. Instead we
+     * scan for the permalink by its value — an instagram.com/p|reel link is a product-level invariant
+     * that barely changes, so the value is a more stable anchor than any obfuscated field name.
+     * Bounded: walk the item's List fields, first element only, no recursion; scoped to xma by caller.
+     */
+    public String xmaReshareLink() {
+        try {
+            Class<?> c = this.mediaClass();
+            for (java.lang.reflect.Field f : c.getDeclaredFields()) {
```
swakwork: Avoid finding fields and methods on runtime

extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/DirectItem.java
```diff
@@ -0,0 +1,293 @@
...
+            }
+        } catch (Exception ignored) {
+        }
+        return null;
+    }
+
+    /** First String field on {@code o} whose value is an instagram.com post/reel permalink, query stripped. */
+    private String firstInstagramPermalink(Object o) {
+        try {
+            for (java.lang.reflect.Field g : o.getClass().getDeclaredFields()) {
```
swakwork: Avoid finding fields and methods on runtime

patches/src/main/kotlin/app/crimera/patches/instagram/misc/settings/SettingsResourcePatch.kt
```diff
@@ -35,6 +35,13 @@ val addSettingsActivityPatch =
                     activity.setAttribute("android:exported", "false")
                     application.appendChild(activity)
                 }
+
+                // Viewer for saved deleted (unsent) direct messages
+                activity = document.createElement("activity")
+                activity.setAttribute("android:name", "app.morphe.extension.instagram.patches.dm.DeletedMessagesActivity")
```
swakwork: These need to be included only if the user includes save deleted messages patch. In that case we can include a separate resource patch for it rather than including with Settings patch

extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/DirectItem.java
vishnuprasad0: hey, sry about this one cleaned it up in the latest commit. the field names aren't found at runtime anymor, they're placeholders directItemEntity rewrites at patch time like the rest of the class.
 take a look

extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/DirectItem.java
vishnuprasad0: same fix here , this is still inside xmaReshareLink() so it uses the same patch time resolved names..
 no runtime field hunting. the instagram.com check is just filtering out non-post xma links (like xma_link)
 updated in the last commit

patches/src/main/kotlin/app/crimera/patches/instagram/misc/settings/SettingsResourcePatch.kt
vishnuprasad0: done ,, pulled these out into their own DeletedMessagesResourcePatch and made it a dependency of SaveDeletedMessagesPatch, so they only get added when that patch is included. settings patch doesn't touch them anymore.. look its ok or not

swakwork:
I finally got time to test it patch. First of all the patch works amazing. I havent dig deep into the code. I just got 2 suggestions.

1. Setting the orientation to Portrait -> Instagram atm supports tablets, with supports landscape as well and some freaks like me would be using on them. So we may not need to lock in portrait unless the UI is broken.

2. Action bar header -> The action bar says "this chat" even when I try to access via Piko Settings.

3. User Id -> Is there any reason why we are using user id instead of user name or group name ? Sure in the DB we can use id as the primary key as name can be modified by the user.

The UI could be more intuitive, but ya, we can discuss or handle it as a future PR.

Thanks for this PR 💌

swakwork:
Is username something that can be achieved?

extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/DirectItem.java
swakwork: Sorry I still don't get it, why do we required to find the element everytime on runtime

vishnuprasad0:
> Is username something that can be achieved?

Hey ,
DirectItem doesn't carry a username field, only item_id, user_id, text, timestamp, hide_in_thread, thread_key. So there's no username to grab at the moment a message is captured.

We get the username separately: a small directory (sender_id → username) that's filled from the thread's participant list via noteThreadUsers(). That hook only runs when the participant list loads/refreshes for a thread — it writes the name once and doesn't refresh after. So if a sender's first message lands before that list loads, the row keeps the raw id; and it won't follow later username changes. That's why it can look like it only worked the first time.

swakwork:
> Hey , DirectItem doesn't carry a username field, only item_id, user_id, text, timestamp, hide_in_thread, thread_key. So there's no username to grab at the moment a message is captured.

I'm pretty sure they would expect user names.

Also does it capture all the messages or only those that are deleted? Either way we may need to add delete data to clear media cache as well as db

vishnuprasad0:
> > Hey , DirectItem doesn't carry a username field, only item_id, user_id, text, timestamp, hide_in_thread, thread_key. So there's no username to grab at the moment a message is captured.
> 
> I'm pretty sure they would expect user names.
> 
> Also does it capture all the messages or only those that are deleted? Either way we may need to add delete data to clear media cache as well as db

1)  yep i agree with youu ....   but Usernames do display , the sender name resolves from the participant directory, then thread title. It only falls back to a raw id in one case: the very first message from a sender before that thread's participant list has been harvested. On an active/opened chat it shows names fine. Did you see the raw id on an active chat, or just on the first-ever message? 

2)  Clear button  already exists as user choice . DeletedMessagesActivity has:
 A "Clear" action in the toolbar (line 96-117) 
  clearSaved(threadId) wipes this chat's rows, or all rows when opened from settings.
  
 
 there's no separate media cache to clear — media isn't downloaded, we only store the CDN/permalink URL as the row content. Clearing the DB row is clearing the media reference tooo

swakwork:
> 1. yep i agree with youu ....   but Usernames do display , the sender name resolves from the participant directory, then thread title. It only falls back to a raw id in one case: the very first message from a sender before that thread's participant list has been harvested. On an active/opened chat it shows names fine. Did you see the raw id on an active chat, or just on the first-ever message?

I got users ids on a existing chat ( 1on1 chat not a group)

> 2. Clear button  already exists as user choice . DeletedMessagesActivity has:
>    A "Clear" action in the toolbar (line 96-117)
>    clearSaved(threadId) wipes this chat's rows, or all rows when opened from settings.
> 
> there's no separate media cache to clear — media isn't downloaded, we only store the CDN/permalink URL as the row content. Clearing the DB row is clearing the media reference tooo

My bad I haven't explored it much

vippium:
@vishnuprasad0 Will it also work for Disappearing Messages or not. Like if someone in chat enables the disappearing messages after 24hr. 

I think it should be merged with it. Bcoz both works in the same way.

swakwork:
Hey, can you test the PR on 439 ?

vishnuprasad0:
> @vishnuprasad0 Will it also work for Disappearing Messages or not. Like if someone in chat enables the disappearing messages after 24hr.
> 
> I think it should be merged with it. Bcoz both works in the same way.

 Partly,
 Vanish mode and the 24h auto-delete I haven't tested. The mechanism should suit them.
 we save the message when it arrives, and the disappearing part only removes Instagram's copy afterwards, so our copy should survive the same way an unsend does.
 I'd keep it out of this PR though

vishnuprasad0:
> Hey, can you test the PR on 439 ?

 Tested on 439, works.

 Needed  fixes to get there..
 voice notes and videos were coming back empty because their field names never got resolved.
  Also moved the thread-id lookup to patch time instead of walking objects at runtime

vippium:
Means so many fixes needed. Maybe they'll be resolved in future.

patches/src/main/kotlin/app/crimera/patches/instagram/misc/dm/saveMessages/SaveDeletedMessagesPatch.kt
```diff
@@ -0,0 +1,234 @@
...
+import app.morphe.patcher.util.smali.ExternalLabel
+import app.morphe.util.getFreeRegisterProvider
+import app.morphe.util.registersUsed
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.instruction.ReferenceInstruction
+import com.android.tools.smali.dexlib2.iface.reference.FieldReference
+import com.android.tools.smali.dexlib2.iface.reference.MethodReference
+
+private const val HOOK_CLASS = "$INTEGRATIONS_PACKAGE/patches/dm/SavedMessagesHook;"
```
swakwork: We can constant for /patches/ as well

patches/src/main/kotlin/app/crimera/utils/Utils.kt
```diff
@@ -144,6 +146,20 @@ fun Fingerprint.changeStringAt(
...
+fun Fingerprint.changeString(
+    sentinel: String,
+    value: String,
+) {
+    val instruction =
+        method.instructions.first {
+            (it.opcode == Opcode.CONST_STRING || it.opcode == Opcode.CONST_STRING_JUMBO) &&
+                it.getReference<StringReference>()?.string == sentinel
+        }
+    val register = (instruction as OneRegisterInstruction).registerA
```
swakwork: Switch to registersused[0]

patches/src/main/kotlin/app/crimera/patches/instagram/misc/dm/saveMessages/SaveDeletedMessagesPatch.kt
vishnuprasad0: done .. using PATCHES_DESCRIPTOR now

patches/src/main/kotlin/app/crimera/utils/Utils.kt
vishnuprasad0: done .. switched to registersUsed[0]

swakwork:
Check the build logs, looks like it still requires some fixed

vishnuprasad0:
> Check the build logs, looks like it still requires some fixed

 yeah the merge dropped the cntxt import  that file went to activity upstream, fixed 
Builds clean  now- check

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/actionbar/ActionBarPatch.java
```diff
@@ -139,7 +139,7 @@ public static void chatActionBarButton(ViewGroup viewGroup) {
             }
 
             if(SettingsStatus.saveDeletedMessages) {
-                Context context = viewGroup.getContext();
+                android.content.Context context = viewGroup.getContext();
```
swakwork: Isn't Context already imported ?

vishnuprasad0: on this branch yep, but upstream changed that import to Activity so it's gone in the merge 😅 
we can just add it back instead of qualifying it ... should i do that  ?

swakwork: Your branch isn't sync with the dev branch. Look at `userProfileActionBarButton` for instance.

vishnuprasad0: > Your branch isn't sync with the dev branch. Look at `userProfileActionBarButton` for instance.

 yeah my bad , that was stale.. merged dev in and retested on 439

AdrianoCahete:
For curiosity: when this change will be on dev patches? I see that the latest tag is from 10/Aug

naoverse06:
dev.8

cyberboh:
@naoverse06 Hi, Beforre release to next dev, please test compatibility to X 12.17.0, I built this branch locally but when patching I got error log https://github.com/crimera/piko/pull/1580#issuecomment-5306325648

vippium:
@swakwork I don't think that the feature is working as expected. Look at this. One from the settings and another one from the chat.

Inside chat, it show nothing. But in settings, it shows that a msg is unsend but doesn't show things properly.

swakwork:
> @swakwork I don't think that the feature is working as expected. Look at this. One from the settings and another one from the chat.
> 
> Inside chat, it show nothing. But in settings, it shows that a msg is unsend but doesn't show things properly.
> 
> 
> 
> 

The last time I check they were working fine but it was very long time ago. And also this is why I hate AI

vippium:
I think this PR needs to be recalled. And a proper rework is needed over it.

swakwork:
> I think this PR needs to be recalled.

How?

vishnuprasad0:
@vippium was the unsent message your own or from the other side? and was the toggle already on when it first arrived?

asking because own messages and anything sent before you enabled it never get captured, so they'd show up exactly like that — blank row. trying to narrow down if it's that or something else

vishnuprasad0:
@vippium @swakwork 
   adding a screenshot from my side works fine 

two things to clear up is ..  it never saves your own messages, only the ones others unsend. and  it picks up usernames from the chat data when it loads, so early on a sender can show as an id instead of a name

 the empty entry you got is a different bug, fix is up in #1728 
 if it still shows up after that plz share a bit more on how you tested

screenshot 2026-08-16 at 7 52 45 PM

PrimePax:
@vishnuprasad0 
Dark screen in deleted chat message list , when Light Theme apply


https://github.com/user-attachments/assets/75909437-a74d-41d2-a0df-d56ec22d130d

PR #1534, opened 2026-07-17
fix(Instagram): Refine Piko settings title styling

naoverse06:
Piko settings screen titles could wrap to a second line with longer translations, while category titles were too large compared with Instagram's settings UI.

Keep screen titles on one line, preserving a 25sp root title and automatically scaling subpage titles from 20sp to 18sp when needed. Use 15sp bold text for preference category titles.

## Testing
- `.\gradlew.bat :patches:clean :patches:buildAndroid --console=plain --no-daemon`

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/SettingsActivity.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/SettingsActivity.java
@@ -47,10 +47,12 @@ public class SettingsActivity extends Activity {
         super.onCreate(bundle);
 
         String displayTitle = null;
+        String fragmentName = null;
 
         // Extract both variables safely from the incoming intent bundle
         if (getIntent() != null && getIntent().getExtras() != null) {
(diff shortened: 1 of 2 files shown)


swakwork:
Can you check if we can hook any of the already  available activity on Instagram, so that we need not create a new activity for it and add it on manifest, in short just like how Twitter settings is working.

I was planning to look into it but unable to find time for it.

naoverse06:
Reusing ModalActivity seems possible, but it needs more work than the Twitter implementation because Instagram initializes its own fragment during onCreate()

swakwork:
Lcom/instagram/base/activity/IgFragmentActivity;

Can you check if this can be used? This seems to be the parent, super class for most of the existing activities

naoverse06:
IgFragmentActivity is abstract and is not declared in the manifest, it cannot be launched directly. ModalActivity already inherits from it through BaseFragmentActivity. Hooking IgFragmentActivity.onCreate() would affect many activities, and returning early from it would only return control to ModalActivity, which would continue its own initialization. We would therefore still need a concrete host and would still need to handle BaseFragmentActivity.A2M(). For these reasons, ModalActivity remains the better choice.

PR #1537, opened 2026-07-17
fix(Instagram): Refresh localized download folder path

naoverse06:
The download folder summary is only populated when the preference screen is created, it stays stale after returning from the system folder picker. the stored document ID can also expose internal identifiers such as `primary:`. refresh the summary when the settings screen resumes and format the path with Android's storage volume label using the device system locale.

## Testing
- `.\gradlew.bat :extensions:instagram:compileReleaseJavaWithJavac :patches:buildAndroid --console=plain`
- Tested on Samsung Galaxy S26 and OPPO Find X9 Pro

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/SettingsActivity.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/SettingsActivity.java
@@ -17,6 +17,7 @@ import android.graphics.Paint;
 import android.graphics.Typeface;
 import android.graphics.drawable.ColorDrawable;
 import android.os.Bundle;
+import android.preference.Preference;
 import android.preference.PreferenceFragment;
 import android.preference.PreferenceManager;
 import android.preference.PreferenceScreen;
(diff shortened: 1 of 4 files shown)


extensions/instagram/src/main/java/app/morphe/extension/instagram/utils/Pref.java
```diff
@@ -218,8 +228,121 @@ public static boolean downloadUsernameFolder() {
         return SharedPref.getBooleanPref(Settings.DOWNLOAD_USERNAME_FOLDER);
     }
 
+    static String formatDownloadPathForDisplay(String storedPath, String resolvedStorageLabel) {
```
swakwork: Move these to shared package may be inside PikoUtils, because the plan is to include the custom download path and custom downloaded to Twitter as well

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/SettingsActivity.java
```diff
@@ -207,6 +209,15 @@ public static class SettingsFragment extends PreferenceFragment {
 
         Context context;
 
+        @Override
+        public void onResume() {
+            super.onResume();
+            Preference downloadPathPreference = findPreference("piko_download_set_path");
```
swakwork: Can we find a way to abstract it, so that we can use it in for other preferences as well like ListPref

PR #1543, opened 2026-07-19
fix(Instagram): Match Piko settings background to Amoled theme patch

naoverse06:
The piko settings screen uses a pure black background in dark mode even when the Amoled theme patch is not applied. use Instagram's `igds_prism_black` resource for the dark mode background instead. this keeps the standard `#0c1014` background without the Amoled theme patch and switches it to pure black only when the patch is applied.

## Testing
- `.\gradlew.bat :extensions:instagram:compileDebugJavaWithJavac --console=plain`

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
@@ -8,7 +8,6 @@ package app.morphe.extension.instagram.settings.preference.widgets;
 
 import android.animation.ValueAnimator;
 import android.content.Context;
-import android.content.res.Configuration;
 import android.graphics.Canvas;
 import android.graphics.Color;
 import android.graphics.Paint;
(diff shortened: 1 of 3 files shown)


extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/widgets/InstagramPreferenceStyle.java
```diff
@@ -51,7 +52,13 @@ public static int dp(Context context, float value) {
     }
 
     public static int backgroundColor() {
-        return UI.getThemedColour("igds_color_primary_background");
+        int primaryBackground = UI.getThemedColour("igds_color_primary_background");
+        int nightMode = Utils.getContext().getResources().getConfiguration().uiMode
```
swakwork: Check if it works with all 3 themes which Instagram (light/dark/system default). I had to remove this before because it wasn't working as expected.

naoverse06: tested all three themes, they seem to be working as expected

swakwork: If it works can you set theme to settings action bar, I had removed it previously.

You can also probably make a method to retrieve theme, in Utils or PikoUtils and use it everywhere wherever theme check is required

naoverse06:
also made the piko settings system bar icons follow the Instagram theme so they don’t blend into the background when the device and instagram themes differ

PR #1545, opened 2026-07-19
fix(Instagram): Restore and sync ghost action bar icons

### LINKED ISSUE #1480: [IG 3.8.0 dev 3] ghost eye icon not appearing in DM
I selected ghost eye icon to appear in DMs and re-launched, but it didn't appear.

Thank you for your work!

naoverse06:
The chat action bar used the user profile value array, causing the ghost icon selection to be saved as the profile info icon. existing ghost icons also only refreshed on the action bar where they were tapped. use the chat-specific value array, migrate the invalid saved chat value, and keep existing ghost icons synchronized across home tab, inbox tab, profile tab, and chat tab.

## Testing
- `.\gradlew.bat :extensions:instagram:compileDebugJavaWithJavac :extensions:instagram:compileReleaseJavaWithJavac`

Closes #1480

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/utils/Pref.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/utils/Pref.java
@@ -7,7 +7,6 @@
 
 package app.morphe.extension.instagram.utils;
 
-import java.util.HashSet;
 import java.util.Set;
 
 import app.morphe.extension.instagram.settings.Settings;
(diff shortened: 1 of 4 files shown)


extensions/instagram/src/main/java/app/morphe/extension/instagram/utils/Pref.java
```diff
@@ -294,7 +295,12 @@ public static Set<String> userProfileActionBarButtons() {
     }
 
     public static Set<String> chatActionBarButtons() {
-        return SharedPref.getSetPref(Settings.ACTION_BAR_CHAT);
+        Set<String> buttons = new HashSet<>(SharedPref.getSetPref(Settings.ACTION_BAR_CHAT));
+        if (buttons.remove(Constants.AB_PROFILE_INFO_ICON)) {
```
swakwork: I don't get this logic

naoverse06: this migrates the value saved by the dev3/dev4 mapping bug. the chat preference displayed ghost mode as the second entry, but it used the profile value array, selecting it saved PROFILE_INFO_ICON instead of GHOST_MODE_ICON. PROFILE_INFO_ICON is not a valid chat action, so this replaces it with GHOST_MODE_ICON and persists the corrected set. the HashSet copy avoids modifying the set returned by SharedPreferences directly

swakwork: The easiest way I feel would be changing the sharedpref key name, instead of checking and setting each time even after its corrected. Let the user set them again

PR #1554, opened 2026-07-20
fix(Instagram): Match story mention dialog to Instagram theme

naoverse06:
The story mention dialog uses the device theme instead of instagram's theme when their light and dark modes differ. its opaque window background can also appear as square corners while the dialog is opening or closing. this change makes the window background transparent and applies instagram's background and foreground colors to the dialog. in dark mode, the background follows instagram's standard color and becomes pure black only when the amoled theme patch is applied.

## Testing
- `.\gradlew.bat :extensions:instagram:compileDebugJavaWithJavac --rerun-tasks --console=plain`
- Tested on Instagram v435.0.0.37.76 with Piko v3.8.0-dev.5

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/story/PeopleCellDialogBox.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/story/PeopleCellDialogBox.java
@@ -8,10 +8,16 @@ package app.morphe.extension.instagram.patches.story;
 
 import android.app.Dialog;
 import android.content.Context;
+import android.graphics.Color;
+import android.graphics.drawable.ColorDrawable;
+import android.graphics.drawable.ShapeDrawable;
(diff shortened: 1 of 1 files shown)


extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/story/PeopleCellDialogBox.java
```diff
@@ -127,15 +144,60 @@ private static View createPeopleListView(@NonNull Context context,
         return scrollView;
     }
 
-    /**
-     * Simple centered "No data" placeholder shown when the list is empty.
-     */
+    private static int resolveDialogBackgroundColor(@NonNull Context context) {
```
swakwork, line 147: Dont we already have isDarkMode for this?

swakwork:
This also covers the rounded edge which I messed up right?

PR #1555, opened 2026-07-20
fix(Twitter): Apply custom sharing domain to copied links

naoverse06:
When the native share menu is disabled, copied post links bypass the custom sharing domain hook and keep using x.com. apply the hook after the share link branches converge so copied links use the configured custom domain regardless of the native share menu setting.

## Testing
- `.\gradlew.bat :patches:check --console=plain`
- Tested on Twitter v12.7.1-release.0 with Piko v3.8.0-dev.5

### DIFF
--- a/patches/src/main/kotlin/app/crimera/patches/twitter/link/handlemodernsharesheetlinks/HandleModernShareSheetLinks.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/twitter/link/handlemodernsharesheetlinks/HandleModernShareSheetLinks.kt
@@ -20,6 +20,7 @@ import app.morphe.util.findFreeRegister
 import app.morphe.util.indexOfFirstInstruction
 import app.morphe.util.registersUsed
 import com.android.tools.smali.dexlib2.Opcode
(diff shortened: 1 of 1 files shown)


patches/src/main/kotlin/app/crimera/patches/twitter/link/handlemodernsharesheetlinks/HandleModernShareSheetLinks.kt
```diff
@@ -97,20 +98,22 @@ val handleModernShareSheetLinks =
 
                 NewShareSheetLinkFingerprint2.method.apply {
                     val firstGoto = indexOfFirstInstruction(Opcode.GOTO)
+                    val linkMergeIndex = getInstruction<BuilderOffsetInstruction>(firstGoto).target.location.index
+                    val linkMergeInstruction = instructions[linkMergeIndex]
```
swakwork: Any reason for using instructions[] instead of getInstruction()?

swakwork: You can also move these to line 107 as they are only used post that. Keeps the flow clean

PR #1567, opened 2026-07-23
feat(Twitter): x-lite patches

crimera:
test it here - https://github.com/crimera/piko-newx/releases


extensions/twitter/src/main/java/app/morphe/extension/twitter/patches/TimelineEntry.java
```diff
@@ -149,5 +149,71 @@ public static List timelineVideos(List videoEnities){
...
+            java.util.List<Object> filteredList = new java.util.ArrayList<>();
+            Iterable<?> iterable = (Iterable<?>) itemsList;
+            
+            for (Object item : iterable) {
+                if (item == null) continue;
+                boolean isAd = false;
+                
+                // 1. Check entryId
+                try {
+                    java.lang.reflect.Method getEntryId = item.getClass().getMethod("getEntryId");
```
swakwork: I don't think the XLite 's entity classes are obfuscated. We can directly create a stub class and use it instead

extensions/twitter/src/main/java/app/morphe/extension/twitter/patches/TimelineEntry.java
```diff
@@ -149,5 +149,71 @@ public static List timelineVideos(List videoEnities){
...
+                    String entryId = (String) getEntryId.invoke(item);
+                    if (entryId != null) {
+                        String lowerEntryId = entryId.toLowerCase(java.util.Locale.ROOT);
+                        if (lowerEntryId.contains("promoted") ||
```
swakwork: Can't we use the already existing filter option for it rather than duplicating it?

swakwork:
Hey, I'm planning to go for a stable release today or tomorrow, as the Dev Branch is stable.
We can probably include this on the next dev release post stable release

crimera:
> Hey, I'm planning to go for a stable release today or tomorrow, as the Dev Branch is stable.
> We can probably include this on the next dev release post stable release 

yeah, all good. not in a rush to get this merged or anything. just wanted to test the waters.

cyberboh:
> Hey, I'm planning to go for a stable release today or tomorrow, as the Dev Branch is stable. We can probably include this on the next dev release post stable release

I don't think so, I upgraded patches from dev.4 to dev.5, then opened Twitter, there were error pop up 'syntetic lambda' (or so) while it load changelogs.
I don't have logs because it appear once, after I open Twitter again, it disappear.
I think error appear only while load changelogs.

swakwork:
> I don't think so, I upgraded patches from dev.4 to dev.5, then opened Twitter, there were error pop up 'syntetic lambda' (or so) while it load changelogs. I don't have logs because it appear once, after I open Twitter again, it disappear. I think error appear only while load changelogs.

Known issue but it doesn't affect the user experience majorly

swakwork:
Adding new activity means, users won't be available mount with root right ?

crimera:
> Adding new activity means, users won't be available mount with root right ?

yeah. tried reusing the old authorizedactivity but the app crashes, with it failing to find the activity. It seems that switching to x-lite disables components from "twitter" https://github.com/crimera/piko/pull/1567/changes/6eeb2b41fd0a1cbb76336c9f4b439b6624f44379#diff-7a9b98256634a2a29f66976a7efe3e7b44ea6ecd47e277157c8c072364e4a3a5R42 . so yeah had to do that workaround for now. will check later if some other "x-lite" activity can be used.

swakwork:
You can merge it if it's complete. It will not trigger a build. We will have to trigger it manually anyways

yoo07:
> Adding new activity means, users won't be available mount with root right ?

I don't think it's a problem, really nobody (or practically nobody) mounts twitter.

patches/src/main/kotlin/app/crimera/patches/twitter/timeline/disableAutoScroll/DisableAutoScrollPatch.kt
```diff
@@ -36,15 +79,40 @@ val disableAutoScrollPatch =
...
+                throw PatchException(
+                    "Expected one X-Lite home reselect handler, found ${homeReselectMatches.size}: " +
+                        homeReselectMatches.joinToString { it.originalMethod.toString() },
+                )
+            }
+
+            homeReselectMatches.single().method.addInstructions(
+                0,
+                """
+                const/4 v0, 0x0
```
swakwork: looks like its not using the preference

cyberboh:
May "For You" tab can be hidden? I did a search for it's settings, but i can't found it.

cyberboh:
today i get new update 12.20.5-prod.01. is it possible to add/port patch `Disable playstore update` to hide update from playstore? Regular piko patch has this patch.

swakwork:
> today i get new update 12.20.5-prod.01. is it possible to add/port patch `Disable playstore update` to hide update from playstore? Regular piko patch has this patch.

You can use `change version code patch` which should be an universal patch (ideally)

PR #1582, opened 2026-07-25
feat(Twitter): Add Dynamic color switch toggle

### LINKED ISSUE #1588: Twitter: Piko setting screen is displayed very difficult to see in light mode
Twitter version: 12.7.1-release.0
Piko version: 3.8.0

Piko setting screen is displayed very difficult to see if the device theme is set as light mode.

Even though the device is set to light mode, the background is displayed in black. In addition, the letters are displayed in white which is almost black, making it very difficult to read the letters.

This problem occurs only on the top screen of the Piko settings, and not on other pages.

Top screen:
Image

Version information screen:
Image

naoverse06:
Adds a dynamic color switch under customize. also fixes the following issues:
- Dynamic color not working correctly with twitter's dim theme
- Piko settings using incorrect colors when twitter is set to use device settings

## Testing
- `.\gradlew.bat :extensions:twitter:compileDebugSources :patches:compileKotlin :patches:checkStringResources --console=plain`
- Tested on Twitter v12.7.1-release.0 with Piko v3.8.0

Closes #1588


extensions/twitter/src/main/java/app/morphe/extension/twitter/settings/SettingsSearchUIController.java
```diff
@@ -9,14 +9,13 @@
 import android.annotation.SuppressLint;
```
swakwork, line 9: Can you use the already available drawables for search icon and clear icon, instead of creating one

extensions/twitter/src/main/java/app/morphe/extension/twitter/patches/customise/appIcon/AppIconManager.java
```diff
@@ -69,26 +74,66 @@ public void disableAllIcons() {
...
+
     // ---------------------------------------------------------------------------------------------
     // Enable ONE icon, disable all others
     // ---------------------------------------------------------------------------------------------
     public void applyIcon(String componentName) {
         String prevComponentName = getSavedIcon();
         saveSelectedIcon(componentName);
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S
+                && app.morphe.extension.twitter.Utils.getBooleanPref(Settings.DYNAMIC_COLOR)
```
swakwork: Call it via Pref

extensions/twitter/src/main/java/app/morphe/extension/twitter/settings/ScreenBuilder.java
```diff
@@ -842,6 +844,21 @@ public void buildCustomiseSection(boolean buildCategory){
...
             );
         }
+
+        if (SettingsStatus.dynamicColor) {
+            Preference dynamicColorPreference = helper.switchPreference(
+                    str("piko_pref_dynamic_color"),
+                    "",
+                    Settings.DYNAMIC_COLOR
+            );
+            dynamicColorPreference.setOnPreferenceChangeListener((preference, newValue) -> {
```
swakwork: Helper class has a on preference change listener. Handle this there

extensions/twitter/src/main/java/app/morphe/extension/twitter/settings/SettingsSearchColors.java
```diff
@@ -5,43 +5,69 @@
...
+        int searchTextColor = Utils.resolveColor(context, "coreColorPrimaryText");
+
+        return new SettingsSearchColors(
+                settingsBackgroundColor,
+                Utils.resolveColor(context, "abstractColorFadedGray"),
+                Utils.resolveColor(context, "coreColorPressed"),
+                Utils.resolveColor(context, "coreColorBorder"),
+                searchTextColor,
+                Utils.resolveColor(context, "coreColorSecondaryText"),
+                withAlpha(
```
swakwork: I'm pretty sure there will be already attributes for handling alpha in resource, check once

patches/src/main/kotlin/app/crimera/patches/twitter/misc/dynamiccolor/DynamicColorPatch.kt
```diff
@@ -1,182 +1,204 @@
...
-                    colorElement.setAttribute("name", k)
-                    colorElement.textContent = v
-
-                    resourcesElement.appendChild(colorElement)
-                }
+                val intValueIndex = instructionMatches[1].index
+                val applyStyleIndex = instructionMatches[2].index
+                val styleRegister =
+                    method
+                        .getInstruction<OneRegisterInstruction>(intValueIndex + 1)
```
swakwork: You can use `getInstructions(<index>).registersUsed[0]` intead

extensions/twitter/src/main/java/app/morphe/extension/twitter/settings/SettingsStatus.java
```diff
@@ -87,7 +87,13 @@ public class SettingsStatus {
     public static boolean customEmojiFont = false;
     public static boolean inlineDownloadButton = false;
 
+    public static boolean dynamicColor = false;
     public static boolean moreInfoOnProfile = false;
+
+    public static void dynamicColor() {
+        dynamicColor = Build.VERSION.SDK_INT >= Build.VERSION_CODES.S;
```
swakwork, line 94: You might also need to check the patch is included

swakwork:
I priority fixes over this. Will include on the next one

patches/src/main/kotlin/app/crimera/patches/twitter/misc/dynamiccolor/DynamicColorPatch.kt
```diff
@@ -1,182 +1,201 @@
...
+    }
 
 @Suppress("unused")
 val dynamicColorPatch =
-    resourcePatch(
+    bytecodePatch(
         name = "Dynamic color",
-        description = "Replaces the default Twitter Blue with the user's Material You palette.",
+        description = "Adds an option to replace Twitter Blue with the user's Material You palette.",
         default = false,
```
kitadai31: Since this patch is no longer permanent, it can be default.

patches/src/main/kotlin/app/crimera/patches/twitter/misc/dynamiccolor/DynamicColorPatch.kt
```diff
@@ -1,182 +1,201 @@
...
-                    FileWriter(stylesXml).use {
-                        it.write("<?xml version=\"1.0\" encoding=\"utf-8\"?><resources></resources>")
-                    }
+            ThemeApplierFingerprint.apply {
+                if (!AccessFlags.STATIC.isSet(method.accessFlags) ||
+                    method.parameters.size != 3 ||
+                    method.parameters.first().type != "Landroid/content/Context;"
+                ) {
+                    throw PatchException("Unexpected Twitter theme applier signature")
                 }
```
kitadai31: Such checks can be performed in the `custom` parameter of the fingerprint.
By doing so, the fingerprint won't match a wrong method in the first place, so there is no need to perform the check on the patch side.

Same goes for ComponentFactoryActivityFingerprint.

kitadai31:
This patch uses a custom app icon patch, but afaik there is a bug with custom app icons in the recent version of X.
The bug is that it causes all icons (activity-aliases) to become enabled, resulting in app drawer getting messed up and cannot launch piko.

Won't this issue occur by dynamic color toggle also changing the app icon?

patches/src/main/kotlin/app/crimera/patches/twitter/misc/dynamiccolor/DynamicColorPatch.kt
```diff
@@ -64,6 +60,11 @@ private object ThemeApplierFingerprint : Fingerprint(
                 returnType = "V",
             ),
         ),
+    custom = { methodDef, _ ->
+        AccessFlags.STATIC.isSet(methodDef.accessFlags) &&
```
swakwork, line 64: I guess you can add access flags as separate parameters which accepts a list.

patches/src/main/kotlin/app/crimera/patches/twitter/misc/dynamiccolor/DynamicColorPatch.kt
```diff
@@ -144,26 +97,16 @@ val dynamicColorPatch =
     bytecodePatch(
         name = "Dynamic color",
         description = "Adds an option to replace Twitter Blue with the user's Material You palette.",
-        default = false,
+        default = true,
```
swakwork, line 100: Set the default to false, many may not want this by default

kitadai31: Since this patch no longer makes permanent resource changes, it can be default
Just like many other patches those have a toggle in Piko settings

patches/src/main/kotlin/app/crimera/patches/twitter/misc/dynamiccolor/DynamicColorPatch.kt
naoverse06: The accessFlags parameter performs an exact comparison with method.accessFlags, while the target method may have additional flags. the custom predicate intentionally checks only whether STATIC is present. it is also needed to express the non-static condition for ComponentFactoryActivityFingerprint

kitadai31:
I fixed change custom icon patch in #1624
You can restore the material you icon process

But rather than linking the app icon to the dynamic color toggle, I think it would be better to add it as a new option to the existing “Change app icon” patch.

Reason:

- Users may want to use custom app icons while also using the dynamic color within the app
- Changing app icon is unstable on some OEM launcher
- Minimize issues when the patch is excluded
  - If dynamic color is enabled in settings and user repatched the app without the dynamic color patch, the activity-alias will be missing and the user cannot launch the app, even if the "Change app icon" patch is included
- A single "Change app icon" patch bear a responsibility for handling the app icon change

If you adopt this approach, I think a guidance for changing app icon can be added to the summary of the toggle.
Like this: `Applies the user's Material You palette to the app-wide theme. To apply dynamic colors to the app icon, go to Piko settings > Customize > Change app icon`

PR #1584, opened 2026-07-26
fix(Instagram): Download max res 4K media

bikram-agarwal:
## Improvements
- Fix Instagram downloads selecting a lower-quality image or video when the first returned variant is not the highest resolution.
- Request image metadata up to 4096px whenever media downloading is enabled, while keeping high-resolution in-app viewing independently controlled by its existing setting.
- Select image and video variants by pixel count, including individual and “Download all media” actions.
- Safely handle empty variant lists instead of attempting to access the first entry.

## 📦 Others
- Verified with `:extensions:instagram:compileDebugJavaWithJavac` and IDE lint checks.

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/MediaData.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/MediaData.java
@@ -252,38 +252,16 @@ public class MediaData extends Entity {
 
     public String getVideoLink() throws Exception {
         List<VideoData> videoDataList = this.getVideoVariants();
-        if(videoDataList != null && !videoDataList.isEmpty()){
-            VideoData highestResolutionVideo = videoDataList.get(0);
-            long highestPixelCount = (long) highestResolutionVideo.getWidth() * highestResolutionVideo.getHeight();
-
-            for (VideoData videoData : videoDataList) {
-                long pixelCount = (long) videoData.getWidth() * videoData.getHeight();
-                if (pixelCount > highestPixelCount) {
-                    highestResolutionVideo = videoData;
-                    highestPixelCount = pixelCount;
-                }
-            }
-
-            return highestResolutionVideo.getUrl();
+        if(videoDataList!=null){
+            return videoDataList.get(0).getUrl();
         }
         return null;
     }
     
     public String getImageLink() throws Exception {
         List<ImageData> imageDataList = this.getImageVariants();
-        if(imageDataList != null && !imageDataList.isEmpty()){
-            ImageData highestResolutionImage = imageDataList.get(0);
-            long highestPixelCount = (long) highestResolutionImage.getWidth() * highestResolutionImage.getHeight();
-
-            for (ImageData imageData : imageDataList) {
-                long pixelCount = (long) imageData.getWidth() * imageData.getHeight();
-                if (pixelCount > highestPixelCount) {
-                    highestResolutionImage = imageData;
-                    highestPixelCount = pixelCount;
-                }
-            }
-
-            return highestResolutionImage.getUrl();
+        if(imageDataList!=null){
+            return imageDataList.get(0).getUrl();
         }
         return null;
     }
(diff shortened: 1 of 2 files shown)


extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/MediaData.java
```diff
@@ -252,16 +252,38 @@ public List getImageVariants() throws Exception {
...
+            return highestResolutionVideo.getUrl();
         }
         return null;
     }
     
     public String getImageLink() throws Exception {
         List<ImageData> imageDataList = this.getImageVariants();
-        if(imageDataList!=null){
-            return imageDataList.get(0).getUrl();
+        if(imageDataList != null && !imageDataList.isEmpty()){
```
swakwork: This compulation is not required. The first one will always have the highest resolution

bikram-agarwal: It isn't working like that. I've encountered it in multiple posts over the weeks. Images I downloaded via Piko patched Insta were low res for some posts while same posts gave high res / 4K images when I downloaded via desktop chrome + Imagus extension. 

After the changes I made in this branch, the patched app is also giving the highest res / 4K for those posts.

swakwork: Setting the Max resolution to 4096 is enough, as the media object list will always be in descending order.
With this changes, turn on Piko debug and check in which order the list is for 4k posts

bikram-agarwal: ahh... verified. Just max res 4K is enough. Pushed the commit.

extensions/instagram/src/main/java/app/morphe/extension/instagram/utils/Pref.java
```diff
@@ -203,11 +204,13 @@ public static String customiseStoryTimestamp() {
     }
 
     public static int improveImageViewing(int defaultSize) {
-        return SharedPref.getBooleanPref(Settings.IMPROVE_IMAGE_VIEWING) ? 2048 : defaultSize;
+        return SharedPref.getBooleanPref(Settings.IMPROVE_IMAGE_VIEWING) ? MAX_IMAGE_SIZE : defaultSize;
     }
 
     public static Integer improveImageViewing(Integer defaultSize) {
-        return SharedPref.getBooleanPref(Settings.IMPROVE_IMAGE_VIEWING) ? 2048 : defaultSize;
+        return enableDownload() || SharedPref.getBooleanPref(Settings.IMPROVE_IMAGE_VIEWING)
```
swakwork: improve image viewing is an whole seperate patch to downloading, we may not need to mix them. Some may not want high resolution media to be fetched

bikram-agarwal: So, should I restore this function to - 

```
    public static Integer improveImageViewing(Integer defaultSize) {
        return SharedPref.getBooleanPref(Settings.IMPROVE_IMAGE_VIEWING) ? MAX_IMAGE_SIZE : defaultSize;
    }
```

So that only when user has explicitly enabled “Improve image viewing”, only then view + download size will max at 4096.

OR, should I add a separate "Download highest quality" toggle in piko settings -> Download options ?

swakwork: There can't be `Download highest quality` per say without `Improve image quality` patch. The download patch doesn't call any API but rather uses the already available data. If user enables the quality patch, they will get the highest quality image data. We can disable quality patch and still get highest resolution. Hope I made sense.

bikram-agarwal: > We can disable quality patch and still get highest resolution. 

I'm guessing this was a typo and you meant "we can't". 

I pushed the new commit that fetches max 4K only when user enables improveImageViewing. So that users who don't enable that don't get 4K fetched unnecessarily.

PR #1585, opened 2026-07-26
feat(Instagram): New patch to block onboarding permission prompts

bikram-agarwal:
## ✨ New Features
- Add a default-enabled `Disable onboarding permission prompts` patch for Instagram.
- Suppress launch-time contacts and location onboarding screens by disabling the `ig_device_permission_consent` and `ig4a_d0_retention` Mobile Config flows.
- Preserve Android permission state, allowing users to grant contacts or location access later when intentionally using related features.

## 📦 Others
- Reuse the existing Mobile Config flag hook and settings patch infrastructure without introducing additional fingerprints.
- Verified by compiling the Instagram extension and Kotlin patches and building the complete MPP bundle.

## 🖼️ Example prompts this patch blocks

Prompt

Issue #1574, opened 2026-07-24
[Instagram] setup new device option keeps reappearing


mian196:
Duplicate of https://github.com/crimera/piko/issues/1279
Piko patches:  3.8.0
Instagram: 435 (suggested version and build number) 

Screenshot: 

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/0a1c28b6-8f79-4f3d-b935-d0ce5f319609" />

I haven't done anything related to location settings or anything i just patched it with latest patches today and this happened on every visit, everytime i open Instagram i see this, is this a bug related ot Instagram patches or i messed up with some Instagram settings?

mian196:
In the original issue it was mentioned it got fixed but im on latest patches this issue still exist

kuntal-devrat:
Try turning off the Disable analytics option and deleting the analytic cache, then force stop the application and reopen the app

mian196:
This works! 
But can i turn on the disable analytics option again? or do i need to make it off until a fix came

kuntal-devrat:
yeah you can disable that afterwards...

I'll solve this in the upcoming patch to fix this issue entirely.😅

mian196:
Yeah i disabled analytics and i see that pop-up again only fix for now is enabling analytics, hopefully this gets fixed

mian196:
This is happening again in latest dev patches  : [v3.9.0-dev.2](https://github.com/crimera/piko/releases/tag/v3.9.0-dev.2)

mian196:
fixed in latest dev patches

Issue #1563, opened 2026-07-22
[Instagram] Downloading media error


AyeeFranklin:
<img width="1440" height="450" alt="Image" src="https://github.com/user-attachments/assets/8c8012ca-efda-44e5-955d-04bbcf0ce7ff" />
For the past week, everytime I try to download photos or videos, it comes with this same error message.

leo328-png:
Same here, I was just about to open a ticket about it. It’s frustrating, and in my case it’s not just that – anything added to the interface gives me this error.

kuntal-devrat:
I'll look into it✌️

LauriLepik:
Same here. Just keep getting this error whenever I click download on any kind of post.

Galaxy Fold7 on Android 16 btw

Xahalya:
Hi, I am having the same issue here, I tried a lot of different things interestingly enough it happened to me once I tried to reinstall the app, before it was doing okay but I had other modification I wanted to add, 

Whatever I do (from activating everything from deactivate one by one, doing a new package name...) I always have this error, what makes me feel weird is that I don't have a prompt consent for storage, tho I am trying to add manually the location in the option menu,

Greetings :)

naoverse06:
Can you reproduce the issue on the latest dev version?

Xahalya:
It depends on what do you mean by "Reproduce" because whatever the build I use I do have this issue, it is not something I can reproduce on itself it is something that happens on the direct functionality, 

I will redownload the lib and retry, I will tell you then :)

Xahalya:
I'm back:

MorpheApp: 1.31.0 (Updated)
Piko (No update available, 3.9.0)
Apkm Instagram: 439.0.0.37.89

Repatched, installation through update (I am not able yet to do a full clean install)

Error still appears as soon as I try to download,

The weird thing is that using pico debug I can fetch the direct server link of the media, it looks like something is happening through the download method itself or as if android security was not providing an auth token to write in the storage? 

I am not so knowledgeable so I cannot tell you more for now :) 

Edit: At this point I am less sure it comes from the patch itself but rather Android messing up, I only had once the query to authorise Instagram to use specific folders and I wonder if there is not something Messing up here, 

If you have any clue to reset auth given by the past to retry it correctly it could be good to test maybe :) 

Have a good day

LauriLepik:
Hello.

[...]

PR #1586, opened 2026-07-26
fix(instagram): pass current carousel index to reel overflow buttons

### LINKED ISSUE #1562: cannot download multiple images in a reel
if a reel has multiple images, downloader only sees the first one

kuntal-devrat:
## Problem

When a reel post contains multiple images (carousel/slideshow), tapping the download button or any other overflow menu button (external downloader, more options) always operated on **slide 1**, regardless of which slide the user was currently viewing. The index was hardcoded to `0` in all `ReelButton` subclasses.

## Root Cause

The reel overflow button hook (`ClipsOrganicMediaItemViewMoreOptionsController`) did not pass the current carousel position to the extension. No index was read from the controller at hook time.

## Fix

The hook now reads `CURRENT_MEDIA_FIELD` from `MEDIA_ADD_INFO_CLASS_NAME` on the controller class — the same field and pattern already used by the feed onClick hook (`HookOverflowMenuButtonOnClickPatch`).

A safe three-instruction register sequence is used to avoid clobbering live registers (which caused a previous crash):

```
iget-object freeReg, self, mediaExtraDataField    ← read extra-data obj (self still intact)
iget        freeReg, freeReg, CURRENT_MEDIA_FIELD ← read int index (safe reuse of freeReg)
iget-object self,    self,    appActivityField    ← clobber self LAST, after all reads done
```

Falls back to index `0` via a 3-arg overload if `mediaExtraDataField` is not found on the controller class.

`hookReelOverflowMenuButton` now depends on `decoderEntity` to ensure `CURRENT_MEDIA_FIELD` and `MEDIA_ADD_INFO_CLASS_NAME` are resolved before the hook runs.

## Testing

Confirmed on a real device — downloading from slide 3 of a carousel reel correctly downloads slide 3. Single-video reels and the reel overflow bottom sheet are unaffected.

## Note

The consent popup fix (#1574, #1563) is handled separately in PR #1585 which introduces a dedicated `Disable onboarding permission prompts` patch.

Closes #1562

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/overflowMenuButton/reels/buttons/DebugButton.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/overflowMenuButton/reels/buttons/DebugButton.java
@@ -14,7 +14,7 @@ import app.morphe.extension.instagram.entity.MediaData;
 
 public class DebugButton extends ReelButton {
     public DebugButton(Context context, Object mediaObject) {
-        super(context, mediaObject);
+        super(context, mediaObject, 0);
     }
 
     @Override
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/overflowMenuButton/reels/buttons/DownloadButton.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/overflowMenuButton/reels/buttons/DownloadButton.java
@@ -12,12 +12,12 @@ import android.content.Context;
 import app.morphe.extension.instagram.patches.download.DownloadUtils;
 
 public class DownloadButton extends ReelButton {
-    public DownloadButton(Context context, Object mediaObject) {
-        super(context, mediaObject);
+    public DownloadButton(Context context, Object mediaObject, int currentMediaIndex) {
+        super(context, mediaObject, currentMediaIndex);
     }
 
     @Override
     public void onClick(View view) {
-        DownloadUtils.downloadPost(this.context, null, this.mediaObject, 0);
+        DownloadUtils.downloadPost(this.context, null, this.mediaObject, this.currentMediaIndex);
     }
 }
--- a/patches/src/main/kotlin/app/crimera/patches/instagram/misc/overflowMenuButton/reels/HookReelOverflowMenuButton.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/instagram/misc/overflowMenuButton/reels/HookReelOverflowMenuButton.kt
@@ -6,6 +6,9 @@
 
 package app.crimera.patches.instagram.misc.overflowMenuButton.reels
 
+import app.crimera.patches.instagram.entity.decoder.CURRENT_MEDIA_FIELD
+import app.crimera.patches.instagram.entity.decoder.MEDIA_ADD_INFO_CLASS_NAME
+import app.crimera.patches.instagram.entity.decoder.decoderEntity
 import app.crimera.patches.instagram.misc.download.AddReelButtonFingerprint
 import app.crimera.patches.instagram.utils.Constants.ADD_REEL_BTN_OVERFLOW_MENU_BUTTON_CLASS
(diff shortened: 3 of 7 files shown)


extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/overflowMenuButton/reels/buttons/DownloadButton.java
```diff
@@ -12,12 +12,16 @@
 import app.morphe.extension.instagram.patches.download.DownloadUtils;
 
 public class DownloadButton extends ReelButton {
+    public DownloadButton(Context context, Object mediaObject, int currentMediaIndex) {
+        super(context, mediaObject, currentMediaIndex);
+    }
+
     public DownloadButton(Context context, Object mediaObject) {
-        super(context, mediaObject);
+        super(context, mediaObject, 0);
```
swakwork: You can remove the 2 parameter constructors, if it's not being used anywhere

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/overflowMenuButton/reels/buttons/DebugButton.java
```diff
@@ -13,8 +13,12 @@
 import app.morphe.extension.instagram.entity.MediaData;
 
 public class DebugButton extends ReelButton {
+    public DebugButton(Context context, Object mediaObject, int currentMediaIndex) {
```
swakwork: Debug button doesn't require the current media index. Let it have 2 parameter constructor with hardcoded current media index as 0

kuntal-devrat:
Addressed — removed the unused 2-parameter constructors from all button subclasses (DownloadButton, ExternalDownloadButton, InfoButton, DebugButton) and from the base ReelButton class. Force pushed.

its good to go

patches/src/main/kotlin/app/crimera/patches/instagram/misc/overflowMenuButton/reels/HookReelOverflowMenuButton.kt
```diff
@@ -22,32 +25,71 @@ val hookReelOverflowMenuButton =
...
+                        iget v$freeRegisterOne, v$freeRegisterOne, $CURRENT_MEDIA_FIELD
+                        iget-object v$selfClassRegister, v$selfClassRegister, $appActivityField
+                        invoke-static {v$selfClassRegister,v$buttonAdderInstanceRegister,v$mediaObjectRegister,v$freeRegisterOne},$ADD_REEL_BTN_OVERFLOW_MENU_BUTTON_CLASS->includeCustomReelOverflowButtons(Landroid/content/Context;Ljava/lang/Object;Ljava/lang/Object;I)V
+                        """.trimIndent(),
+                    )
+                } else {
+                    // Fallback: reel controller doesn't have MEDIA_ADD_INFO field.
+                    // Pass index 0 — safe, single-item reels are always index 0.
+                    addInstructions(
+                        mediaObjectFromParameterIndex + 1,
```
swakwork, line 71: Is the fallback really necessary?

kuntal-devrat:
Addressed all three comments:

1. **DebugButton** — reverted to 2-param constructor with hardcoded index 0, since it doesn't use the index.
2. **Fallback else branch** — removed. The controller class confirmed to always have the \MEDIA_ADD_INFO_CLASS_NAME\ field (tested on device). The fallback was dead code.
3. Removed the now-unused 3-arg \includeCustomReelOverflowButtons\ overload from \AddReelButton\ as well.

Force pushed.

swakwork:
Just noticed MEDIA_ADD_INFO_CLASS_NAME isn't present in the Add Reels Button class

kuntal-devrat:
> Just noticed MEDIA_ADD_INFO_CLASS_NAME isn't present in the Add Reels Button class

My bad 😔 and I saw that you fixed that.

I will be more careful from now on 😅

Actually I'm new into this....so please

PR #1596, opened 2026-07-27
fix(instagram): patch GRAY_1600 to fix Compose surfaces staying gray in AMOLED mode

ViictorNunes:
AMOLED theme only overrode colors.xml, but newer Compose screens read
their dark surface color from GRAY_1600, a compiled constant inside
BasePrismColorsV2's <clinit> - not a resource, so those overrides never
reached it, leaving some screens dark gray instead of black.

Added a bytecode patch that finds BasePrismColorsV2's <clinit> via a
Fingerprint and rewrites the constant to pure black (0xff000000L).
Wired into the Theme patch via dependsOn.

Credit: instafel (https://github.com/mamiiblt/instafel) for identifying
GRAY_1600 as the missing piece.

### DIFF
--- a/patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ComposePrismBlackPatch.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ComposePrismBlackPatch.kt
@@ -1,0 +1,68 @@
+/*
+ * Copyright (C) 2026 piko <https://github.com/crimera/piko>
+ *
+ * See the included NOTICE file for GPLv3 §7(b) terms that apply to this code.
+ */
+
+package app.crimera.patches.instagram.misc.theme
+
+import app.crimera.patches.instagram.utils.Constants.COMPATIBILITY_INSTAGRAM
+import app.morphe.patcher.Fingerprint
+import app.morphe.patcher.InstructionLocation
+import app.morphe.patcher.extensions.InstructionExtensions.replaceInstruction
+import app.morphe.patcher.fieldAccess
+import app.morphe.patcher.opcode
+import app.morphe.patcher.patch.bytecodePatch
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
+
+// Thanks to instafel (https://github.com/mamiiblt/instafel) - GRAY_1600 is a
+// compiled constant in BasePrismColorsV2's <clinit>, not a color resource, so
+// the colors.xml overrides in ThemePatch.kt don't reach it. It's the
+// prism-black tone newer Compose screens use, so without this patch they stay
+// dark gray instead of pure black in AMOLED mode.
+private const val PRISM_COLORS_V2_CLASS = "Lcom/instagram/compose/core/theme/BasePrismColorsV2;"
+private const val GRAY_1600_FIELD_NAME = "GRAY_1600"
+private const val PURE_BLACK_LITERAL = "0xff000000L"
+
+// Matches the const-wide that feeds GRAY_1600's sput-wide directly: a
+// shl-long/2addr sits in between (packing the ARGB value into Compose's
+// internal 64-bit color representation), so MatchAfterWithin(1) allows for
+// exactly that one unmatched instruction and no more - this is what pins the
+// const-wide to the one immediately preceding GRAY_1600 specifically, instead
+// of matching the first const-wide in the method.
+internal object ComposePrismBlackFingerprint : Fingerprint(
+    definingClass = PRISM_COLORS_V2_CLASS,
+    name = "<clinit>",
+    filters = listOf(
+        opcode(Opcode.CONST_WIDE),
+        fieldAccess(
+            definingClass = PRISM_COLORS_V2_CLASS,
+            name = GRAY_1600_FIELD_NAME,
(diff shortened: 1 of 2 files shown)


vippium:
Finally someone looked into it. Thnx votsz

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemePatch.kt
```diff
@@ -37,12 +46,64 @@ val themePatch =
...
+private const val PURE_BLACK_LITERAL = "0xff000000L"
+
+internal object BasePrismColorsV2ClinitFingerprint : Fingerprint(
+    custom = { method, classDef ->
+        classDef.type == PRISM_COLORS_V2_CLASS && method.name == "<clinit>"
+    },
+)
+
+context(context: BytecodePatchContext)
+private fun patchGray1600(): Boolean {
```
swakwork: Create this to a separate Bytecode patch and add that as the dependency to the existing resourcepatch

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemePatch.kt
```diff
@@ -37,12 +46,64 @@ val themePatch =
...
+    BasePrismColorsV2ClinitFingerprint.method.apply {
+        val sputIndex = instructions.indexOfFirst { instruction ->
+            instruction.opcode == Opcode.SPUT_WIDE &&
```
swakwork: You can use built in field extractor and method extractor functions

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemePatch.kt
```diff
@@ -37,12 +46,64 @@ val themePatch =
...
+                (instruction as? ReferenceInstruction)?.reference
+                    ?.let { it as? FieldReference }
+                    ?.let { it.definingClass == PRISM_COLORS_V2_CLASS && it.name == GRAY_1600_FIELD_NAME } == true
+        }
+
+        if (sputIndex == -1) return false
+
+        for (index in sputIndex - 1 downTo 0) {
```
swakwork: Use filter instead

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemePatch.kt
```diff
@@ -37,12 +46,64 @@ val themePatch =
...
+            val instruction = instructions[index]
+
+            if (instruction.opcode == Opcode.CONST_WIDE) {
+                val register = (instruction as OneRegisterInstruction).registerA
```
swakwork: You can use .registeredUsed[] instead

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemePatch.kt
```diff
@@ -37,12 +46,64 @@ val themePatch =
...
+                replaceInstruction(index, "const-wide v$register, $PURE_BLACK_LITERAL")
```
swakwork: Is it hardcoding to black or will it also support dynamic theming

ViictorNunes: The Material You theme doesn't edit Instagram's existing colors directly — it builds its own `piko_dynamic_*` alias palette, then remaps Instagram's real color tokens to point at those aliases.

The alias palette exists in 4 resource folders with identical names but different values: `res/values` and `res/values-night` hold fixed hex fallbacks for pre-Android-12 devices (light/dark), while `res/values-v31` and `res/values-night-v31` point at the system's dynamic (wallpaper-based) palette for Android 12+. `prism_black/prism_white` are treated as mode-invariant constants — black stays black and white stays white in both light and dark, unlike every other role which flips.

`materialYouNamedMappings` then repoints Instagram's actual color tokens (e.g.` igds_color_primary_button`) to these aliases across all 4 folders, so the dynamic palette propagates through the whole app without touching each screen individually. `ensureColorsXml()` creates the `-v31` folders/files first since they don't exist in the stock APK.

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ComposePrismBlackPatch.kt
```diff
@@ -0,0 +1,68 @@
...
+            name = GRAY_1600_FIELD_NAME,
+            opcode = Opcode.SPUT_WIDE,
+            location = InstructionLocation.MatchAfterWithin(1),
+        ),
+    ),
+)
+
+internal val composePrismBlackPatch =
+    bytecodePatch(
+        name = "Fix Compose prism black",
```
swakwork: Don't give it a name. If name parameter is there, morphe will treat it as a seperate patch, which isn't the case here

ViictorNunes: ok

johnnyw94:
Hi dev @ViictorNunes , thanks for this. But is it possible to apply this to some sections which this doesn't apply in 'Settings and activity' for fully immersive amoled?

PR #1600, opened 2026-07-27
fix(Instagram): Resolve bytecode verifier crashes

### LINKED ISSUE #1592: Instagram crashes when trying to open message section
Instagram crashes on message and profile section. But works fine with search/home/reels. 
Piko version 3.9.0 dev-1

naoverse06:
Fixes multiple issues occurring in instagram 439.0.0.37.89 patched with piko 3.9.0-dev.1. the inbox crash occurred because the action bar hook ran before `CHECK_CAST` established the register as an `IgdsActionBar`, causing the verifier to reject a `View` passed as a `ViewGroup`. the profile crash occurred because the friendship status patch skipped a required integer initialization, leaving a register typed as a `String` where an integer was expected. moves the inbox hook after the required type conversion and preserves the integer initialization before bypassing the internal badge checks. restores the profile action bar hook after instagram's action bar builder and validates the target method, required fields, parameters, and registers before injection. the profile action bar dependency and settings fallback behavior are also restored.

## Testing
- `.\gradlew.bat :patches:check --rerun-tasks --console=plain`
- Tested on Instagram 439.0.0.37.89 with Piko 3.9.0-dev.1

Closes #1592 , #1601

### DIFF
--- a/patches/src/main/kotlin/app/crimera/patches/instagram/misc/actionBar/inboxActionBarButton/InboxActionBarButtonPatch.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/instagram/misc/actionBar/inboxActionBarButton/InboxActionBarButtonPatch.kt
@@ -40,7 +40,7 @@ val inboxActionBarButtonPatch =
                     if (typeRef == "Lcom/instagram/igds/components/actionbar/IgdsActionBar;") {
                         val viewGroupRegister = it.registersUsed[0]
                         addInstructions(
-                            it.location.index,
+                            it.location.index + 1,
                             """
                             invoke-static {v$viewGroupRegister}, $ACTIONBAR_DESCRIPTOR->inboxActionBarButton(Landroid/view/ViewGroup;)V
                             """.trimIndent(),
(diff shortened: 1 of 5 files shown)


patches/src/main/kotlin/app/crimera/patches/instagram/misc/actionBar/inboxActionBarButton/InboxActionBarButtonPatch.kt
```diff
@@ -35,21 +36,26 @@ val inboxActionBarButtonPatch =
         execute {
 
             InboxActionBarBuilderFingerprint.method.apply {
-                instructions.filter { it.opcode == Opcode.CHECK_CAST }.firstOrNull {
-                    val typeRef = it.getReference<TypeReference>()!!.type
-                    if (typeRef == "Lcom/instagram/igds/components/actionbar/IgdsActionBar;") {
-                        val viewGroupRegister = it.registersUsed[0]
-                        addInstructions(
-                            it.location.index,
```
swakwork, line 43: I was about to push the fix, thanks.

But I feel like the rewrite may not be required, simply adding + 1 here should do the job

swakwork:
Is it possible to separate the profile action bar fix as a different commit as it's a different fix. If it's not possible no worries

swakwork:
Good to merge ?

PR #1602, opened 2026-07-27
fix(instagram): inject inboxActionBarButton after CHECK_CAST to fix VerifyError on 439

kuntal-devrat:
## Problem

On Instagram **439.0.0.37.89** (Android 16 / SDK 36), the app crashes immediately when navigating to Direct Inbox:

\\\
java.lang.VerifyError: Verifier rejected class X.08PF:
  android.view.View X.08PF.onCreateView(...) failed to verify:
  register v9 has type Reference: android.view.View
  but expected Reference: android.view.ViewGroup
\\\

## Root cause

\InboxActionBarButtonPatch\ injects the \inboxActionBarButton\ call at \it.location.index\ — **before** the \CHECK_CAST to IgdsActionBar\ executes. At that point the register still holds the raw \ndroid.view.View\ returned by \LayoutInflater.inflate()\.

The DEX verifier (ART, SDK 36) sees that register being passed where a \ViewGroup\ is declared, and rejects the entire \X.08PF\ (DirectInboxFragment) class.

## Fix

Inject at \index + 1\ (after the cast), so the register is already typed as \IgdsActionBar\ — a \ViewGroup\ subclass — when the verifier inspects it.

This is exactly the same pattern \ChatActionBarButtonPatch\ already uses (with an explicit comment explaining why).

## Tested on
- Instagram \439.0.0.37.89\ / Android 16 / SDK 36 (iQOO device)


swakwork:
Thanks but it's already addressed https://github.com/crimera/piko/pull/1600

Issue #1572, opened 2026-07-24
[Twitter] 3.8.0 some patches don't work on version 12.7.1-release.0


mhcp00:
I tested the latest version 3.8.0 release on the version of twitter 12.7.1-release.0, Morphe manager says everything is fine, no errors, no warings.
Patches "Show sensitive media" and "Remove premium upsell" don't work.

Multisquid2k3:
I have the same issue

FateJapan:
I am facing the same problem.

LandonMoran:
Me too

ax1121:
I would like to add that trying to verify my age by scanning a selfie from the internet isn't working either.

mhcp00:
3.9.0-dev.1 resolve the issue "Remove premium upsell" while my PR #1605 fix "Show sensitive media"

PR #1603, opened 2026-07-28
fix(twitter): fix show sensitive media #1572

mhcp00:


mhcp00:
sorry, wrong branch source

PR #1605, opened 2026-07-28
fix(twitter): show sensitive media #1572

mhcp00:

### DIFF
--- a/patches/src/main/kotlin/app/crimera/patches/twitter/timeline/sensitivemediasettings/ShowSensitiveMediaPatch.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/twitter/timeline/sensitivemediasettings/ShowSensitiveMediaPatch.kt
@@ -8,20 +8,22 @@ package app.crimera.patches.twitter.timeline.sensitivemediasettings
 
 import app.crimera.patches.twitter.misc.settings.settingsPatch
 import app.crimera.patches.twitter.utils.Constants.COMPATIBILITY_X
-import app.crimera.patches.twitter.utils.Constants.PATCHES_DESCRIPTOR
+import app.crimera.patches.twitter.utils.Constants.PREF_DESCRIPTOR
 import app.crimera.patches.twitter.utils.enableSettings
 import app.morphe.patcher.Fingerprint
-import app.morphe.patcher.extensions.InstructionExtensions.addInstructions
+import app.morphe.patcher.extensions.InstructionExtensions.addInstructionsWithLabels
 import app.morphe.patcher.extensions.InstructionExtensions.instructions
 import app.morphe.patcher.patch.bytecodePatch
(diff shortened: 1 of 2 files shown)


mhcp00:
Tested on version 12.11

patches/src/main/kotlin/app/crimera/patches/twitter/timeline/sensitivemediasettings/ShowSensitiveMediaPatch.kt
```diff
@@ -33,19 +35,28 @@ val sensitiveMediaPatch =
...
-            val TIMELINE_ENTRY_DESCRIPTOR = "$PATCHES_DESCRIPTOR/TimelineEntry"
-
-            val methods = sensitiveMediaSettingsPatchFingerprint.method
-            val instructions = methods.instructions
-
-            val returnObj = instructions.last { it.opcode == Opcode.RETURN_OBJECT }.location.index
-
-            methods.addInstructions(
-                returnObj,
+            val method = sensitiveMediaSettingsPatchFingerprint.method
```
swakwork, line 38: I feel like it would be better if we hook the parse method like we have done for timeline items and handle the fields on java rather than in smali.

mhcp00: Which fields? 
There are no fields to edit but just avoid loading a part of the JSON. 
The old fields are no longer used by the current version of X. 
Much of the code is left but is not being executed.

mhcp00: I can restore the code inside TimelineEntry and change the sensitiveMedia method so that it returns to a boolean and then inside the JsonTweetWithVisibilityResults$$JsonObjectMapper class, verify that boolean to enable or not the loading of media_visibility_results from TweetWithVisibilityResults.

swakwork: Take a look at how ads from TimelineEntry is handled. Basically pass the `Lcom/twitter/model/json/core/JsonTweetWithVisibilityResults;` to Java hook and handle the checks there and return back the updated result object back

mhcp00:
Too complicated in my opinion, I prefer my version, which is simple.
Good luck with your project.

swakwork:
No worries, if you are okay, can I update on your work?

BlueStarBRS:
Pls both Things are anoying as hell Without a fix thx 🙏🏻

mhcp00:
new PR #1620 I hope it's good

mhcp00:
> can I update on your work?

sure, I think the new PR meets your specifications

PR #1620, opened 2026-07-30
fix(twitter): show sensitive media #1572

mhcp00:

I hope this version is good for @swakwork 

However I don't like it, Java Reflection is slow, I tried to optimize it as much as possible.

### DIFF
--- a/extensions/twitter/src/main/java/app/morphe/extension/twitter/patches/TimelineEntry.java
+++ b/extensions/twitter/src/main/java/app/morphe/extension/twitter/patches/TimelineEntry.java
@@ -20,7 +20,7 @@ import app.morphe.extension.crimera.PikoUtils;
 public class TimelineEntry {
     public static final boolean hideAds;
     private static final boolean hideWTF,hideCTS,hideCTJ,hideDetailedPosts,hideRBMK,hidePinnedPosts,hidePremiumPrompt,showSensitiveMedia,hideTopPeopleSearch,hideTodaysNews;
-    private static final Field mediaVisibility;
+    private static Field mediaVisibility;
 
     static {
         hideAds = (Pref.hideAds() && SettingsStatus.hideAds);
@@ -34,7 +34,6 @@ public class TimelineEntry {
         showSensitiveMedia = Pref.showSensitiveMedia();
         hideTopPeopleSearch = (Pref.hideTopPeopleSearch() && SettingsStatus.hideTopPeopleSearch);
         hideTodaysNews = (Pref.hideTodaysNews() && SettingsStatus.hideTodaysNews);
-        mediaVisibility = getFieldMediaVisibility();
     }
 
     private static boolean isEntryIdRemove(String entryId) {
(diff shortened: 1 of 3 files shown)


mhcp00:
Tested also on latest 12.11.1

extensions/twitter/src/main/java/app/morphe/extension/twitter/patches/TimelineEntry.java
```diff
@@ -136,7 +138,7 @@ public static List timelineVideos(List videoEnities){
                     maxVideoObject = vidObj;
                 }
                 if (maxVideoObject != null) {
-                    ArrayList result = new ArrayList();
+                    ArrayList<Object> result = new ArrayList<>();
```
mhcp00, line 144: this change is for fix a warning in the compiler, I like it when the compiler doesn't complain.

extensions/twitter/src/main/java/app/morphe/extension/twitter/patches/TimelineEntry.java
```diff
@@ -148,6 +150,15 @@ public static List timelineVideos(List videoEnities){
 
         return videoEnities;
     }
+    private static final Field getFieldMediaVisibility() {
+        Field[] fields = JsonTweetWithVisibilityResults.class.getDeclaredFields();
 
+        for (Field field : fields) {
+            if (field.getType().getName().indexOf("model.mediavisibility") != -1) {
```
swakwork: You can get the field name while patching instead of runtime

mhcp00: I saw this class ./extensions/twitter/src/main/java/app/morphe/extension/twitter/patches/TweetInfo.java
I'll try to catch it during patching

mhcp00: while patching does it mean that I then have to pass it as a parameter to the sensitiveMedia method inside TimelineEntry?

swakwork: No, create a function that returns a String and make it return a dummy string. And while patching update the string with the field name.
Just like how entity classes are built

https://github.com/crimera/piko/blob/88762b9e5462622d426e0d282266a29d358c0309/patches/src/main/kotlin/app/crimera/patches/twitter/entity/tweet/TweetEntityPatch.kt#L98

zh89ale: @swakwork given your fussiness in accepting this PR, could you tell me why this code exists in this class?
https://github.com/crimera/piko/blob/88762b9e5462622d426e0d282266a29d358c0309/extensions/twitter/src/main/java/app/morphe/extension/twitter/patches/TweetInfo.java#L36-L48

Wouldn't it be better to put a hook in the JSON reception and modify it, adding or removing parts as needed? There is already a hook for logging ( Server-Response), just need to put a method there that checks what is enabled and what is not.  
In my opinion it would be faster, what is the point of parsing the JSON and then deleting the object?

swakwork: Some of the code are legacy code, which hasn't been touched in ages. If you dig deep, you can find plenty such opportunities for refactoring.
Feel free to point them out

PR #1625, opened 2026-07-31
fix(instagram): handle null comment media and user data when saving comment media

kuntal-devrat:
### Summary
Fixes potential NullPointerExceptions and reflection failure crashes when attempting to save comment media on comments where user details or media metadata are missing or unparseable.

### Changes
- Wrapped getImageMedia() and getGifMedia() field getters in CommentData.java with try-catch blocks to safely return 
ull when field reflection fails.
- Added null safety checks for userData and userData.getUsername() in HandleCommentButton.java with fallback timestamp naming.
- Wrapped comment media download handler in exception handling with toast feedback.

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/CommentData.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/CommentData.java
@@ -42,18 +42,25 @@ public class CommentData extends Entity implements GifMediaInterface {
     }
 
     public MediaData getImageMedia() throws Exception {
-        Entity imageDataEntity = super.getFieldAsEntity("fieldName");
-        if (imageDataEntity != null) {
-            Object mediaObj = imageDataEntity.getField("fieldName2");
-            if (mediaObj != null) {
-                return new MediaData(mediaObj);
+        try {
+            Entity imageDataEntity = super.getFieldAsEntity("fieldName");
+            if (imageDataEntity != null) {
+                Object mediaObj = imageDataEntity.getField("fieldName2");
+                if (mediaObj != null) {
+                    return new MediaData(mediaObj);
+                }
             }
+        } catch (Exception ignored) {
         }
         return null;
     }
 
     private Object getGifMedia() throws Exception {
-        return super.getField("fieldName");
+        try {
+            return super.getField("fieldName");
+        } catch (Exception ignored) {
+            return null;
+        }
     }
 
     public boolean hasGifMedia() throws Exception {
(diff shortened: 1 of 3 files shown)


gradle.properties
```diff
@@ -2,4 +2,4 @@ org.gradle.parallel = true
 org.gradle.caching = true
 kotlin.code.style = official
 android.useAndroidX = true
-version = 3.9.0-dev.2
+version = 3.9.0-dev.3
```
swakwork, line 5: Dont update the gradle files

extensions/instagram/src/main/java/app/morphe/extension/instagram/entity/CommentData.java
```diff
@@ -42,18 +42,25 @@ public boolean hasImageMedia() throws Exception {
...
+        } catch (Exception ignored) {
         }
         return null;
     }
 
     private Object getGifMedia() throws Exception {
-        return super.getField("fieldName");
+        try {
+            return super.getField("fieldName");
+        } catch (Exception ignored) {
```
swakwork, line 61: I feel it's useless.
We are already throwing the exception and catching the exception on the handler, then why

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/comment/HandleCommentButton.java
```diff
@@ -89,15 +90,20 @@ public static boolean checkOnCommentButtonClick(Object button, List list) {
                         MediaData imageData = commentData.getImageMedia();
                         UserData userData = commentData.getCommentUserData();
 
-                        String userName = (userData != null) ? userData.getUsername() : "comment_"+LocalDateTime.now().format(DateTimeFormatter.ofPattern("yyyyMMdd_HHmmss"));
+                        String userName = (userData != null && userData.getUsername() != null) ? userData.getUsername() : "comment_" + System.currentTimeMillis();
```
swakwork, line 93: `yyyyMMdd_HHmmss` is the pattern we are following every where else

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/comment/HandleCommentButton.java
```diff
@@ -89,15 +90,20 @@ public static boolean checkOnCommentButtonClick(Object button, List list) {
...
                         String mediaLink = imageData.getMediaLink();
                         String fileName = userName + "_" + imageData.getDownloadFilename(MediaType.ANY);
 
                         String subFolder = DownloadUtils.getSubfolderName(userName);
 
                         DownloadUtils.downloadMediaUrl(context, mediaLink, subFolder, fileName);
+                    } else {
+                        PikoUtils.toast(str("piko_comment_copied_failed"));
```
swakwork, line 101: Why do you want to toast "copy failed" when an user is downloading something

swakwork:
Things looks good as it is how. This PR doesn't add any significance according to me

PR #1637, opened 2026-08-01
fix(Instagram): resolve the biography accessor, not the translated one

tjelite1986:
## Problem

`Copy bio` on a profile returns nothing, and any patch reading `UserData.getBio()` gets an empty string.

The placeholder in the extension *is* rewritten — the failure is which method it is rewritten to. `BiographyLiveTreeUserDictFingerprint` asks for the method carrying the string `biography`, but a string match is a substring match, and `LiveTreeUserDict` also holds `translated_biography` and `has_biography_translation`. What resolves is the translated-biography getter:

```
com.instagram.user.model.LiveTreeUserDict.DT4:()Ljava/lang/String;
  iget-object v2, v3, LiveTreeUserDict;.A00:LX/2en;
  iget-object v0, v2, LX/2en;.A7x:Ljava/lang/String;
  if-nez v0, ...
  const-string/jumbo v0, "translated_biography"
  invoke-interface {v1, v0}, Ljava/util/Set;.contains:(Ljava/lang/String;)Z
  ...getOptionalStringValueNative...
```

That returns null for any account whose bio was never translated, i.e. nearly all of them — and null is a perfectly ordinary answer for "no bio", so nothing throws and nothing is logged.

## Fix

Pin the return type to `String` and require an instruction whose string reference is exactly `biography`.

## Verification

Disassembling the patched APK (Instagram 435.0.0.37.76):

- before: `getBio()` carries `const-string v2, "DT4"`
- after: `getBio()` carries `const-string v0, "BG2"`

`BG2()Ljava/lang/String;` is the accessor that reads the `biography` key. Copy bio works on device after the change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


patches/src/main/kotlin/app/crimera/patches/instagram/entity/userdata/Fingerprint.kt
```diff
@@ -95,13 +97,16 @@ internal object FriendshipStatusLiveTreeUserDictFingerprint : Fingerprint(
 internal object BiographyLiveTreeUserDictFingerprint : Fingerprint(
     strings = listOf("biography"),
     definingClass = LIVE_TREE_USER_DICT_CLASS,
+    returnType = "Ljava/lang/String;",
+    // "biography" also occurs inside "translated_biography" and
```
swakwork: Comments may not be required as its self explanatory

kitadai31:
You can use instruction filter `string` instead of the `strings = ` parameter or custom block
The string filter uses exact matching of `StringComparisonType.EQUALS` by default, so you don't need to use custom block

```
filters =
        listOf(
            string("biography"),
        ),
```

swakwork:
> You can use instruction filter `string` instead of the `strings = ` parameter or custom block The string filter uses exact matching of `StringComparisonType.EQUALS`, so you don't need to use custom block
> 
> ```
> filters =
>         listOf(
>             string("biography"),
>         ),
> ```

LOL thanks, I was wondering what did I do wrong, that it was not matching the string

PR #1638, opened 2026-08-01
feat(Instagram): fill out the profile options menu

tjelite1986:
## What

The profile options menu can already download a profile picture, copy the username, the full name, the user id and the bio. This adds the neighbours that were missing:

- **View profile picture** — opens the picture it could already download
- **Copy profile link** — `https://www.instagram.com/<username>/`
- **Share this profile** — the same link through the system share sheet
- **Copy links from bio** — the URLs written in the bio, one per line

## Notes

The link matcher is loose on purpose: bios write `example.com` far more often than `https://example.com`, so a scheme is optional and added when the link is handed out. Duplicates are dropped so a repeated link does not read as noise, and a bio with no links says so instead of copying an empty clipboard.

Five new strings, no new settings, nothing removed. Builds against Instagram 435.0.0.37.76.

Depends on nothing else, but note that `Copy links from bio` only has text to work with once #1637 is in — `getBio()` currently resolves to the translated-biography accessor and returns null for almost every account.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/userprofile/ProfileMoreOption.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/userprofile/ProfileMoreOption.java
@@ -13,6 +13,8 @@ import android.app.Dialog;
 import android.content.DialogInterface;
 import android.view.ViewGroup;
 import java.util.ArrayList;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import app.morphe.extension.shared.Logger;
 import app.morphe.extension.shared.Utils;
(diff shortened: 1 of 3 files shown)


extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/userprofile/ProfileMoreOption.java
```diff
@@ -110,6 +139,33 @@ public void onClick(DialogInterface d, int which) {
         }
     }
 
+    /** The public permalink of a profile: what "share" and "copy link" both hand out. */
+    private static String profileLink(String username) {
```
swakwork: You can probably now this inside userdata class such that we can simply call `userdata.profileLink()`  to get it

tjelite1986: Done in 316c6192 — `UserData` now builds the permalink from the username it already owns, the private helper in `ProfileMoreOption` is gone, and both call sites just ask for it.

I named it `getProfileLink()` rather than the bare `profileLink()` you wrote, to match `getUsername()` / `getFullName()` / `getUserId()` / `getBio()` on the same class. Happy to rename if you prefer it without the prefix.

PR #1639, opened 2026-08-02
fix(Instagram): fix Material You toggle title and crash after accessibility settings redesign

ViictorNunes:
The `clips closed captions h1 2026 redesign accessibility global setting` moved Instagram's Accessibility
screen to a new Compose-based fragment (AccessibilityOptionsComposeFragment),
which changed nullability and resource-resolution behavior enough to expose
existing bugs in the Material You/AMOLED toggle that the old screen never hit.

- Add MaterialYouTheme.getMaterialYouTitle(), mirroring the existing
  hardcoded getAmoledTitle(), since neither string is translated
- Legacy switch (Android 8-11) now calls getMaterialYouTitle() instead of
  IgStr.str("piko_material_you_title"), which resolved to an unrelated
  native string ("No, thanks") since that resource name was never backed
  by a real value, only a reserved public id
- Compose toggle (Android 12+) now calls getMaterialYouTitle() instead of
  embedding that same unset resource id directly
- Compose toggle description is set to an empty string instead of null,
  fixing a NullPointerException on IG 439.0.0+ where the redesigned row
  builder now requires a non-null description
- Legacy AMOLED radio reuses an existing native title resource id instead
  of reserving its own, since the displayed text is always overridden by
  getLegacyRadioTitle() regardless of that id's value
- Remove the now-unused piko_material_you_title/piko_amoled_title public
  string reservation and strings.xml entries. They aren't necessary because there is no translation for those two words.
  
  `Obviously, it was much easier to disable the flag, but I realized there was a fix for it.`
  
  
image

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/theme/MaterialYouTheme.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/theme/MaterialYouTheme.java
@@ -17,7 +17,7 @@ import android.os.Looper;
 import android.widget.CompoundButton;
 import android.widget.RadioGroup;
 
-import app.morphe.extension.crimera.SharedPref;
+import app.morphe.extension.crimera.sharedPreference.SharedPref;
 import app.morphe.extension.instagram.settings.Settings;
 import app.morphe.extension.instagram.utils.IgStr;
 import app.morphe.extension.shared.Logger;
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/theme/MaterialYouTheme.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/theme/MaterialYouTheme.java
@@ -17,7 +17,7 @@ import android.os.Looper;
 import android.widget.CompoundButton;
 import android.widget.RadioGroup;
 
-import app.morphe.extension.crimera.SharedPref;
+import app.morphe.extension.shared.SharedPref;
 import app.morphe.extension.instagram.settings.Settings;
 import app.morphe.extension.instagram.utils.IgStr;
 import app.morphe.extension.shared.Logger;
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/theme/MaterialYouTheme.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/theme/MaterialYouTheme.java
@@ -17,7 +17,7 @@ import android.os.Looper;
 import android.widget.CompoundButton;
 import android.widget.RadioGroup;
 
-import app.morphe.extension.shared.SharedPref;
+import app.morphe.extension.crimera.SharedPref;
 import app.morphe.extension.instagram.settings.Settings;
 import app.morphe.extension.instagram.utils.IgStr;
 import app.morphe.extension.shared.Logger;
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/theme/MaterialYouTheme.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/theme/MaterialYouTheme.java
@@ -17,7 +17,7 @@ import android.os.Looper;
 import android.widget.CompoundButton;
 import android.widget.RadioGroup;
 
-import app.morphe.extension.crimera.sharedPreference.SharedPref;
+import app.morphe.extension.crimera.SharedPref;
 import app.morphe.extension.instagram.settings.Settings;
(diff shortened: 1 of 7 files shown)


patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemeLegacySwitchBytecode.kt
swakwork, line 362: Dont hardcode the registers unless its added at the start of the function

extensions/instagram/src/main/java/app/morphe/extension/instagram/theme/MaterialYouTheme.java
```diff
@@ -17,7 +17,7 @@
 import android.widget.CompoundButton;
 import android.widget.RadioGroup;
 
-import app.morphe.extension.crimera.sharedPreference.SharedPref;
+import app.morphe.extension.crimera.SharedPref;
```
swakwork: `SharedPref` package name got updated on the newer versions. Please align with it

ViictorNunes: I didn't understand, it's to leave as it is or revert to as I had initially put it `app.morphe.extension.crimera.sharedPreference.SharedPref` ?

swakwork: `app.morphe.extension.crimera.sharedPreference.SharedPref` is the new package name

swakwork:
I feel like the simplest approach would be having it in piko settings, rather than Instagram's. It does look cool but would be hard to maintain.

There are some build errors as well you can probably take a look at it

ViictorNunes:
> I feel like the simplest approach would be having it in piko settings, rather than Instagram's. It does look cool but would be hard to maintain.
> 
> There are some build errors as well you can probably take a look at it

Regarding moving to Piko settings, I don't know how that would work, but it's a good idea.

I'll take a look later.

swakwork:
> Regarding moving to Piko settings, I don't know how that would work, but it's a good idea.

Its just a toggle right ? or are we depended on any of the values from that page ?

patches/src/main/kotlin/app/crimera/patches/instagram/misc/theme/ThemeLegacySwitchBytecode.kt
```diff
@@ -174,6 +174,162 @@ private fun deriveSwitchItemBinding(): LegacySwitchItemBinding {
...
+    """
+    invoke-static {}, Lapp/morphe/extension/instagram/theme/MaterialYouTheme;->getLegacyToggleListener()Landroid/widget/CompoundButton${'$'}OnCheckedChangeListener;
+    move-result-object v1
+    invoke-static {}, Lapp/morphe/extension/instagram/theme/MaterialYouTheme;->getMaterialYouTitle()Ljava/lang/String;
+    move-result-object v2
+    invoke-static {}, Lapp/morphe/extension/instagram/theme/MaterialYouTheme;->isEnabled()Z
+    move-result v3
+    new-instance v4, $switchType
+    invoke-direct {v4, v1, v2, v3}, $switchConstructor
+    invoke-interface {v6, v4}, Ljava/util/Collection;->add(Ljava/lang/Object;)Z
```
swakwork, line 206: Don hardcode registers

ViictorNunes:
> > Regarding moving to Piko settings, I don't know how that would work, but it's a good idea.
> 
> Its just a toggle right ? or are we depended on any of the values from that page ?

I don't know. I believe we should focus on and solve this mini bug and then we can see the probability of moving to the Piko settings.

swakwork:
> I believe we should focus on and solve this mini bug 

Which is?

> and then we can see the probability of moving to the Piko settings.

I still don't get you, why do we need to have that option under the theme section?  Does it have any significance? If all that toggle does is turn the preference on and off,  I don't think we need to have those on that page, moving to piko settings is the right option

ViictorNunes:
> Which is?

Click on the AMOLED theme using an enabled flag to leave all accessibility options exposed. Having it active, it ends up crashing when clicking on the theme. 
 
> I still don't get you, why do we need to have that option under the theme section? Does it have any significance? If all that toggle does is turn the preference on and off, I don't think we need to have those on that page, moving to piko settings is the right option

if you are going to move to Piko settings, this PR is very useless in reality, since it would only serve a flag that conflicts

PR #1643, opened 2026-08-02
fix(Twitter): Align compatibility name with changelog scope

naoverse06:
Morphe manager compares changelog scopes with a patch's compatibility name when deciding whether an update is available. piko's X compatibility name was X, while its existing changelog entries use Twitter. because the names did not match, manager could fail to show piko updates on the home screen. rename the compatibility entry to Twitter and enable bring back twitter patch by default. this aligns the compatibility name, initial app label, and changelog scope, morphe manager can find matching updates


patches/src/main/kotlin/app/crimera/patches/twitter/misc/bringbacktwitter/BringBackTwitterPatch.kt
```diff
@@ -22,7 +22,7 @@ val bringBackTwitterPatch =
     resourcePatch(
         name = "Bring back twitter",
         description = "Bring back old twitter logo and name",
-        default = false,
+        default = true,
```
swakwork: Let it stay off by default

PR #1646, opened 2026-08-02
fix(Instagram): Use native switches in piko settings

naoverse06:
Replaces the custom-drawn switches in piko settings with instagram's native `IgdsSwitch` and animation

## Testing
- Tested on Instagram 439.0.0.37.89 with Piko 3.9.0-dev.2


swakwork:
Will have to check, in the future, if `IgdsSwitch` can be made into a entity class just like `IgdsButton`

PR #1651, opened 2026-08-03
feat(Instagram): Added some recommended flags

ViictorNunes:


swakwork:
Does all the flags requires the entire config enabled/disabled or were you not aware of `::`?
If not then you can control individual sub flags using `::` like `9999::23` but make sure the sub flag id is correct.

ViictorNunes:
> Does all the flags requires the entire config enabled/disabled or were you not aware of `::`? If not then you can control individual sub flags using `::` like `9999::23` but make sure the sub flag id is correct.

I already knew, so I only implemented flags that activate completely.

PR #1652, opened 2026-08-04
fix(Twitter): Restore patch-time dynamic color

naoverse06:
Restores dynamic color as a patch-time option instead of a piko settings toggle

## Testing
- `.\gradlew.bat :extensions:twitter:compileDebugSources :patches:compileKotlin :patches:checkStringResources --console=plain`
- Tested on Twitter v12.11.0-release.0 with Piko v3.9.0-dev.4


swakwork:
Not a fan of custom arsc's. What's its purpose?

naoverse06:
they allow direct twitter blue resource references to follow dynamic color at runtime and return to their original colors when the setting is disabled, theme styles alone do not cover these references

swakwork:
> they allow direct twitter blue resource references to follow dynamic color at runtime and return to their original colors when the setting is disabled, theme styles alone do not cover these references

Hmmm not a fan, but anyways we might sunset existing twitter patches ones XLite is ready

naoverse06:
i’m thinking of reverting it to a patch-time option

PR #1654, opened 2026-08-04
feat(Instagram): add master switch to enable/disable recommended flags

ViictorNunes:
## Problem
Restoring an old preferences backup could leave community-recommended
flags in a state that no longer matched the current recommended list,
with no easy way to fix it without toggling each flag individually.

## Solution
Added a master switch at the top of the Recommended Flags screen.
Off by default: no recommended flags are applied, regardless of their
individual saved state. On: recommended flags are applied as usual,
respecting each individual toggle.

The switch is stored in the main settings file, separate from the
recommended flags file, so it's unaffected by community flags backups.

## Changes
- `Settings.java`: new `ENABLE_RECOMMENDED_FLAGS` setting (default `false`)
- `Pref.java`: new `enableRecommendedFlags()` getter
- `HookFlags.java`: `addRecommendedFlags()` early-returns when disabled
- `ScreenBuilder.java`: new switch preference in `buildRecommendedFlagsSection()`
- `strings.xml`: new strings for the switch title/description

image

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/HookFlags.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/HookFlags.java
@@ -78,6 +78,7 @@ public class HookFlags {
     }
 
     private static void addRecommendedFlags(){
+        if (!Pref.enableRecommendedFlags()) return;
         Map<String, Boolean> recFlags = FlagsSharedPref.getAll();
         BOOL_FLAGS.putAll(recFlags);
     }
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/Settings.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/Settings.java
@@ -17,6 +17,7 @@ public class Settings {
     // It should be true always. It will be handled upon opening the piko settings for the first time.
     public static final BooleanSetting FIRST_TIME_PIKO = new BooleanSetting("first_time_piko", true);
     public static final StringSetting LAST_RECOMMENDED_FLAG_DOWNLOAD_TIMESTAMP = new StringSetting("last_recommended_flag_download_timestamp", "0");
+    public static final BooleanSetting ENABLE_RECOMMENDED_FLAGS = new BooleanSetting("enable_recommended_flags", false);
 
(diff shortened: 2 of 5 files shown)


swakwork:
Sorry, but that's not how they are intended to work. We don't need a master key for it.

ViictorNunes:
> Sorry, but that's not how they are intended to work. We don't need a master key for it.

In my opinion, it should stay. I notice that many users are already utilizing backups with some of the recommended commands I added enabled. If a user is using a backup where any of these commands are enabled or disabled, they will remain that way according to the list. If the user wants that feature back, they need to go to the recommended flags and toggle the specific flag. To avoid this issue, I came up with a "master key" idea: it would only modify the flags if the user explicitly enables the feature and selects the specific flags they want from the recommended list.

kitadai31:
I submitted another feature request: #1657
How about this?
I think this is a better approach than the master switch solution

swakwork:
> In my opinion, it should stay. I notice that many users are already utilizing backups with some of the recommended commands I added enabled. If a user is using a backup where any of these commands are enabled or disabled, they will remain that way according to the list. If the user wants that feature back, they need to go to the recommended flags and toggle the specific flag. 

To avoid one or two flag, why should someone disable the entirety of the recommendation.
If you want me to think from your shoes, I do rather say make recommended flags into a patch, so that people who want it can patch it other can ignore the patch, instead of a universal or master switch

ViictorNunes:
> To avoid one or two flag, why should someone disable the entirety of the recommendation. 

There are crazy people out there, at one time or another it was going to happen.

 >>I submitted another feature request: #1657

Since you didn't like the idea of Master Switch, I liked this idea that I wouldn't need a Master Switch and that I wouldn't force the flag according to the list.

kitadai31:
According to [a comment from Swak](https://github.com/crimera/piko/issues/1657#issuecomment-5195326381), it seems a bit difficult to add a "default" state to the recommended flags...

Given the situation, I think adding the master switch OR make it a new patch are also fine.

swakwork:
> According to [a comment from Swak](https://github.com/crimera/piko/issues/1657#issuecomment-5195326381), it seems a bit difficult to add a "default" state to the recommended flags.

I do am checking if there are any alternative solutions to it

PR #1664, opened 2026-08-08
refactor(Instagram): Recommended flags

swakwork:
* Made recommended flags into a patch
* Switched from boolean switch to list, adding support for default state

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/FlagsSharedPref.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/FlagsSharedPref.java
@@ -12,8 +12,9 @@ import java.util.Iterator;
 import java.util.Map;
 import java.util.HashMap;
 
+import app.morphe.extension.crimera.PikoUtils;
 import app.morphe.extension.crimera.sharedPreference.BaseSharedPref;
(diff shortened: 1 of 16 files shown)


swakwork:
@naoverse06 @kitadai31 thoughts on it ?

naoverse06:
lgtm

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/FlagsSharedPref.java
```diff
@@ -41,7 +42,10 @@ public static Map<String, Boolean> getAll(){
             Iterator<String> keys = flags.keys();
             while (keys.hasNext()) {
                 String key = keys.next();
-                Boolean value = (Boolean) flags.get(key);
+                String flagState = (String) flags.get(key);
+                if (flagState.equals(FlagState.DISABLE.toString())) continue;
```
kitadai31, line 46: Isn't it should be `DEFAULT` ?
It appears that the flag is not overridden when set to DISABLE, but is overridden when set to DEFAULT.

Sorry for reviewing late!

swakwork: sh!t

PR #1671, opened 2026-08-09
fix(Instagram): Improve piko settings handling

naoverse06:
Some piko settings cache their values while the instagram process remains alive. Removing instagram from recents does not always stop the process, saved changes may not be applied until instagram is force stopped.
After a setting is changed, removing instagram from recents now flushes the piko preferences before stopping the remaining process

## Testing
- Tested on Instagram v439.0.0.37.89 with Piko v3.9.0-dev.4


extensions/instagram/src/main/java/app/morphe/extension/instagram/utils/InstaUtils.java
```diff
@@ -27,9 +34,139 @@
...
+    }
+
+    private static void flushPreferences() {
+        Context context = Utils.getContext();
+        if (context == null) {
+            throw new IllegalStateException("Extension context is unavailable");
+        }
+
+        boolean settingsFlushed = context
+                .getSharedPreferences(ExtensionStrings.PIKO_SETTINGS, Context.MODE_PRIVATE)
```
swakwork: You may need to move this to BasedSharedPref and call it via SharedPref

extensions/instagram/src/main/java/app/morphe/extension/instagram/utils/InstaUtils.java
```diff
@@ -27,9 +34,139 @@
...
+        }
+    }
+
+    private static void scheduleProcessExit() {
+        new Handler(Looper.getMainLooper()).post(
+                () -> Process.killProcess(Process.myPid())
+        );
+    }
+
+    private static final class SettingsChangeCoordinator {
```
swakwork: Why do we need to create inner class? If it can be then please create it as individual classes

extensions/instagram/src/main/java/app/morphe/extension/instagram/utils/InstaUtils.java
```diff
@@ -27,9 +34,139 @@
 import app.morphe.extension.shared.requests.Requester;
 import app.morphe.extension.shared.requests.Route;
 import app.morphe.extension.shared.Logger;
-import static app.morphe.extension.shared.requests.Route.Method.GET;
 
 public class InstaUtils {
+    private static final String SETTINGS_RESTART_TAG = "PikoSettingsRestart";
```
swakwork: Can move the entire flushing logic to a separate class. InstaUtils is ideally for actions that are repeatly used.

PR #1674, opened 2026-08-10
perf(Instagram - Disable analytics): Turn off by default, state the side effect to description

kitadai31:
Change the default value for the "Disable analytics" setting to off.

This feature is not a fundamental one that everyone wants like hiding ads, so it does not need to be enabled by default.
And this feature has a known issue that are difficult to fix. (An onboarding popup requesting permission for location and contacts keeps reappearing even after users have allowed or skipped it)
So, it would be best to make it optional, opt-in.

Moreover, this approach actually mitigates that issue.
It is caused by information that the popup has already been viewed is not sent to server, so it does not occur if the user views the dialog, chooses to skip or allow it, and then turns on "Disable analytics."
By turning this off by default, users will only enable it after seeing the onboarding screen once.
Therefore, this change is also useful for people who actually want to block analytics.


CzechHek:
What does patch stopping contacts and location prompts do then? It doesn't work. Why keep it? Why change default disable analytics state and not fix the core issue?

Issue #1648, opened 2026-08-03
bug(Instagram): Buttons are not visible when using amoled theme with material


vippium:
<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/05ec7ccf-b18d-4678-9518-579b16b2ef41" />

There is an issue where some elements have issues with theme coloring. Whenever I use amoled with Material, the buttons gets non visible. They acts as they're still living in light theme, rather than dark theme. 

I think material theme with amoled needs some resource fixes.

vippium:
<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/03d3b647-2af4-4ef0-81a1-cb512d1e54d3" />

vippium:
@swakwork please look into this.

crimera:
https://github.com/crimera/piko/pull/1681

PR #1679, opened 2026-08-11
fix(Instagram): Preserve amoled across theme changes

naoverse06:
Keeps the amoled preference enabled when instagram temporarily switches to a light theme, the light theme uses the no amoled palette while the setting remains selected and unavailable, then automatically restores amoled when instagram returns to dark mode

## Testing
- `.\gradlew.bat clean --no-daemon --console=plain`
- `.\gradlew.bat test :patches:buildAndroid --no-daemon --console=plain`
- Tested on Instagram v439.0.0.37.89 with Piko v3.9.0-dev.7

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/theme/MaterialYouState.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/theme/MaterialYouState.java
@@ -66,37 +66,19 @@ final class MaterialYouState {
         return ((long) overrideArgb) << 32;
     }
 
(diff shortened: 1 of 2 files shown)


vippium:
Will this fix the bug having light theme buttons after selecting material and amoled at same time i.e. material amoled theme ? 
Here is the issue for better understanding:
https://github.com/crimera/piko/issues/1648

johnnyw94:
Will it resolve this issue? 

Amoled not applied fully on the DM search bar, the rectangle should be in amoled right? (Using material theme here)

PR #1681, opened 2026-08-11
fix(Instagram): Preserve reels media control colors

### LINKED ISSUE #1648: bug(Instagram): Buttons are not visible when using amoled theme with material

naoverse06:
Closes #1648


vippium:
@naoverse06 can you also look into 1 more thing related to material theming. In Myinsta, stories rings also gets adapted with material theme. So, if this can also be done, then it'll be a + for material theming.

PR #1686, opened 2026-08-12
fix(Instagram): block permission onboarding screens

Pelotrio:
## What changed

- Block the contact-import and location-services NDX Bloks screens at the full-screen opener.
- Match only the two observed onboarding app IDs, leaving unrelated Bloks screens unchanged.
- Keep the existing experiment-flag and legacy consent-URI protections in place.
- Update the patch description because these prompts can appear during normal use, not only on launch.

## Root cause

On Instagram 439, these prompts can arrive through a server-driven quick-promotion action instead of the consent URI paths already handled by the patch:

```text
QuickPromotionSurfaceQueryV3
-> com.bloks.www.qp.async.bloks_action
-> Bloks full-screen opener
-> BKDataFetcher.fetch
-> ModalActivity
```

The resulting screens use these app IDs:

```text
com.bloks.www.bloks.ig.ndx.ci.entry.screen
com.bloks.www.bloks.ig.ndx.ls.entry.screen
```

The existing flags and URI interception therefore do not cover this entry route. Returning later in the fetch path is also too late because modal navigation may already have been scheduled, leaving an indefinitely loading screen.

The new guard runs at the full-screen opener entry and returns before the data fetch or modal navigation begins.

## User impact

The random contacts/location setup screens are suppressed while other Bloks screens continue normally.

## Validation

- `./gradlew :patches:build --stacktrace`
- Patched Instagram `439.0.0.37.89` (`384510827`, arm64-v8a): all patching, rebuilding, and signing steps succeeded; 55 patches applied and none failed.
- Captured both NDX app IDs from natural prompt occurrences and confirmed they use the same full-screen opener.
- Used a temporary diagnostic build that showed a toast only after the exact production predicate matched. A natural contacts prompt produced the confirmation toast while the onboarding screen did not open.
- Built and installed a final artifact without the diagnostic patch, log tags, or toast strings; Instagram launched without an AndroidRuntime crash.

### DIFF
--- a/patches/src/main/kotlin/app/crimera/patches/instagram/misc/onboarding/DisableOnboardingPermissionPromptsPatch.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/instagram/misc/onboarding/DisableOnboardingPermissionPromptsPatch.kt
@@ -9,15 +9,35 @@ package app.crimera.patches.instagram.misc.onboarding
 import app.crimera.patches.instagram.links.interceptUriPatch
 import app.crimera.patches.instagram.misc.settings.settingsPatch
 import app.crimera.patches.instagram.utils.Constants.COMPATIBILITY_INSTAGRAM
+import app.crimera.patches.instagram.utils.Constants.LINKS_DESCRIPTOR
 import app.crimera.patches.instagram.utils.addFlags
 import app.crimera.patches.instagram.utils.enableSettings
+import app.morphe.patcher.Fingerprint
+import app.morphe.patcher.extensions.InstructionExtensions.addInstructionsWithLabels
+import app.morphe.patcher.extensions.InstructionExtensions.getInstruction
+import app.morphe.patcher.extensions.InstructionExtensions.instructions
 import app.morphe.patcher.patch.bytecodePatch
+import app.morphe.patcher.util.smali.ExternalLabel
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.instruction.ReferenceInstruction
+import com.android.tools.smali.dexlib2.iface.reference.FieldReference
+
+private object IgBloksFullScreenOpenFingerprint : Fingerprint(
+    returnType = "V",
+    parameters =
+        listOf(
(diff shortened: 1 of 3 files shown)


patches/src/main/kotlin/app/crimera/patches/instagram/misc/onboarding/DisableOnboardingPermissionPromptsPatch.kt
```diff
@@ -11,13 +11,34 @@ import app.crimera.patches.instagram.misc.settings.settingsPatch
...
+import app.morphe.patcher.extensions.InstructionExtensions.getInstruction
+import app.morphe.patcher.extensions.InstructionExtensions.instructions
 import app.morphe.patcher.patch.bytecodePatch
+import app.morphe.patcher.util.smali.ExternalLabel
+import com.android.tools.smali.dexlib2.AccessFlags
+import com.android.tools.smali.dexlib2.Opcode
+import com.android.tools.smali.dexlib2.iface.instruction.ReferenceInstruction
+import com.android.tools.smali.dexlib2.iface.reference.FieldReference
+
+private const val LINKS_DESCRIPTOR = "Lapp/morphe/extension/instagram/patches/Links;"
```
swakwork: Use the patches constant to target patches package. Check constants

patches/src/main/kotlin/app/crimera/patches/instagram/misc/onboarding/DisableOnboardingPermissionPromptsPatch.kt
```diff
@@ -11,13 +11,34 @@ import app.crimera.patches.instagram.misc.settings.settingsPatch
...
+        ),
+    strings = listOf("BKDataFetcher.fetch"),
+)
 
 @Suppress("unused")
 val disableOnboardingPermissionPromptsPatch =
     bytecodePatch(
         name = "Disable onboarding permission prompts",
-        description = "Prevents contacts and location permission onboarding prompts from appearing on launch.",
+        description = "Prevents contacts and location permission onboarding prompts from appearing.",
```
swakwork, line 40: Can you also update the strings for disable analytics patch

PR #1698, opened 2026-08-14
fix(Instagram): Correct shared link sanitization

### LINKED ISSUE #1682: bug(instagram): fix sanitize url to remove "igsi" param
Currently the param "igsi" isn't removed.

naoverse06:
Fixes share links being sanitized even when the setting is turned off. when enabled, it removes `igsi` and other unwanted parameters without breaking comments, multi-item posts, or stories, and now also covers highlight and audio share links

## Testing
- `.\gradlew.bat test --no-daemon --console=plain`
- Tested on Instagram v439.0.0.37.89 with Piko v3.9.0-dev.7

Closes #1682

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/Links.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/Links.java
@@ -10,16 +10,15 @@ package app.morphe.extension.instagram.patches;
 import android.net.Uri;
 import java.io.IOException;
 import java.net.URI;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Arrays;
-import java.util.Locale;
 
 import app.morphe.extension.instagram.entity.Entity;
 import app.morphe.extension.instagram.entity.MediaData;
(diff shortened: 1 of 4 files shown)


extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/Links.java
```diff
@@ -155,18 +157,100 @@ public static void interceptUri(URI uri) throws IOException{
...
-                    .replaceAll("([&?])fbclid=[^&]*", "")
-                    .replaceAll("([&?])si=[^&]*", "");
+            return ShareLinkSanitizer.sanitize(url, Pref.sanitizeShareLinks());
         } catch (Exception e) {
             Logger.printException(() -> "sanitizeUrl failed: ", e);
         }
         return url;
     }
 
+    static final class ShareLinkSanitizer {
```
swakwork: as always make it into a separate class. For a matter of fact this can be even moved to the shared package but not just restricting to Instagram

PR #1702, opened 2026-08-14
fix(instagram): colorize FriendshipStatusIndicator text and improve pill contrast/spacing

ViictorNunes:
Match indicator text color in FriendshipStatusIndicator to the follow-back
status color instead of always using the themed secondary text color, so
"Does not follow you" / "Follows you" / "Following each other" are colored
red/green/blue respectively (falling back to the secondary color when the
status color preference is disabled).

Increase the pill background opacity and add a subtle stroke so the
badge is visible in both light and dark mode, instead of being nearly
invisible at 5% alpha.

Add a bottom margin below the badge so it no longer sits flush against
the profile bio text underneath it.

image

image

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/userprofile/FriendshipStatusIndicator.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/userprofile/FriendshipStatusIndicator.java
@@ -100,18 +100,19 @@ public class FriendshipStatusIndicator {
             friendshipStatusTextView.setText(text);
 
             int secondaryTextColor = UI.getThemedColour("igds_color_secondary_text");
-            friendshipStatusTextView.setTextColor(secondaryTextColor);
 
             int indicatorColor = indicatorColorHex == null
                     ? secondaryTextColor
                     : Color.parseColor(indicatorColorHex);
+
+            friendshipStatusTextView.setTextColor(indicatorColor);
+
             Drawable statusIcon = ResourceUtils
                     .getDrawable(indicatorIconDrawable)
                     .mutate();
             statusIcon.setColorFilter(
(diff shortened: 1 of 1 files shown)


naoverse06:
I prefer keeping the text in the themed secondary color and using the status color only for the icon. coloring both the icon and text makes the indicator too prominent, especially in the unfollowed state, and the fixed colors do not have enough text contrast in the light theme. the 2dp stroke and increased vertical padding also make the pill feel too heavy. i think we should keep the current subtle styling and only consider the additional spacing below the indicator

ViictorNunes:
> I prefer keeping the text in the themed secondary color and using the status color only for the icon. coloring both the icon and text makes the indicator too prominent, especially in the unfollowed state, and the fixed colors do not have enough text contrast in the light theme. the 2dp stroke and increased vertical padding also make the pill feel too heavy. i think we should keep the current subtle styling and only consider the additional spacing below the indicator


I understand the concern, but I prefer to keep the color in the text. I just wanted to comment on the points you raised:

About contrast in the light theme - the colors I used ('#EB4941', '#3CC176', '#3389DF') are already used elsewhere in the app to indicate status, so I think the contrast is ok in most cases. If you have a specific example where it was difficult to read, tell me to adjust the tone of that color specifically, instead of taking the whole color.

About being very flashy in the state "doesn't follow" - in fact I think that's exactly where it makes the most sense to draw attention, because it's the information that the user will most likely want to act on (follow back or stop following).

Now, about stroke and padding, I agree with you - that made the pill too heavy. I will take the 2dp stroke and return the vertical padding closer to the original, to keep light even with the colored text.

The spacing underneath I will keep, this is just a visual breath, it does not leave more heavy.

If you still prefer to reverse the color of the text, calm, I just wanted to explain the reasoning first.

PR #1715, opened 2026-08-15
feat(Instagram): Add `Hide online status` patch

pemaismais:
Adds a `Hide online status` patch to the Ghost section: the user stops being shown as active while still seeing everyone else's activity status.

Instagram's own "Show activity status" toggle can't do this. It is bilateral — `is_presence_enabled` gates both the writes and the reads, so turning it off blinds you too.

### How presence actually leaves the app

Three independent channels, all of them silenced here:

| Channel | Where | What the patch does |
|---|---|---|
| DGW/UPI | `PresenceWriteRequest` constructor | rewrites `ACTIVE`/`IDLE` to `OFFLINE` |
| MQTT connection inference | `RealtimeClientManager.createMqttClient` | always subscribes to `/disable_presence_reporting` |
| Legacy MQTT | `RealtimeClientManager$updateAppStateInternal$1.run` | clears the 2nd argument of `HET`, skipping the `/t_fs` publish |

### Notes on the approach

**Why the constructor and not the `publish()` call sites.** The write that opens a session travels as an argument of `IgDgwPresenceClientImpl.establishStream`, and that call is what opens the stream the friends' presence arrives on. Skipping the write would take the reads down with it. Hooking the constructor also covers all three outgoing call sites (`LX/6uW` setup and teardown, `LX/8zp` transitions) with one fingerprint.

**Why `OFFLINE` and `DISABLED` pass through untouched.** Those come from the teardown path, so they are writes worth keeping. The `PresenceStatus` `<clinit>` maps them as `A06 = OFFLINE`, `A03 = ACTIVE`, `A05 = IDLE`, `A04 = DISABLED`.

**Why the enum is resolved by name.** R8 renames the static fields between versions but keeps the enum names as `const-string`, so the extension uses `name()` and `valueOf("OFFLINE")` rather than referencing `->A06`.

**Why `HET` and not `XplatNativeClientWrapper.setForeground`.** With the second argument false, the client calls `setForeground(z, null, null)`, and the null payload makes `setForeground` skip the publish on its own. This is the path the app already takes whenever the UPI rollout flag is on, so `setForegroundNative()` and the `isForeground` bookkeeping keep working.

**What is deliberately not patched.** The RTI legacy client (`LX/kl9;->HET`) ignores its second argument, and its `/t_fs` payload carries the MQTT subscribe/unsubscribe lists alongside the foreground state, so dropping that publish would break realtime subscriptions. It only matters if `MqttClientSelector` picks `kl9` over `8aE`.

Nothing touches `is_presence_enabled`, the `direct_v2/get_presence/` polling or the DGW read stream.

### Testing

Built and applied with `morphe-desktop 1.13.0` against Instagram 435.0.0.37.76 (arm64-v8a), run on an ARM64 emulator. All three fingerprints resolve, and the toggle behaves as described.

That run was built before this branch was rebased onto `dev`, so I am re-verifying the build on this base. Nothing in the patch logic changed in the rebase, only a neighbouring constant in `Constants.kt`.

Not yet covered: repeated foreground/background cycles, MQTT reconnection, force close, and long sessions.

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/Presence.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/Presence.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright (C) 2026 piko <https://github.com/crimera/piko>
+ *
+ * See the included NOTICE file for GPLv3 §7(b) terms that apply to this code.
+*/
+
+
+package app.morphe.extension.instagram.patches;
+
+import com.facebook.presence.model.upi.PresenceStatus;
+
+import app.morphe.extension.instagram.utils.Pref;
+import app.morphe.extension.shared.Logger;
+
+@SuppressWarnings("unused")
+public class Presence {
+    private static final String ACTIVE = "ACTIVE";
+    private static final String IDLE = "IDLE";
+    private static final String OFFLINE = "OFFLINE";
+
+    /**
+     * Called from the PresenceWriteRequest constructor, which every outgoing presence write
+     * goes through.
+     *
+     * <p>The status is rewritten instead of the write being skipped because the first write of
+     * a session travels as an argument of IgDgwPresenceClientImpl.establishStream, and that
+     * call is what opens the stream the friends' presence arrives on.
+     *
+     * <p>OFFLINE and DISABLED are returned untouched: they come from the teardown path, and
+     * those writes are worth keeping.
+     *
+     * <p>The constants are compared by enum name rather than by field because R8 renames the
+     * fields (A03..A06 on 435.0.0.37.76) but keeps the names.
+     */
+    public static PresenceStatus overridePresenceStatus(PresenceStatus original) {
+        try {
+            if (original == null || !Pref.hideOnlineStatus()) {
+                return original;
+            }
+
+            String name = original.name();
+            if (ACTIVE.equals(name) || IDLE.equals(name)) {
+                return PresenceStatus.valueOf(OFFLINE);
+            }
+        } catch (Exception ex) {
+            Logger.printException(() -> "overridePresenceStatus failed: ", ex);
+        }
+        return original;
+    }
+
+    /**
+     * Decides whether "/disable_presence_reporting" goes into the MQTT subscription list, which
+     * stops the server from inferring presence from the connection itself.
+     *
+     * @param original the app's own decision, the UPI rollout flag DA7.A03
+     */
+    public static boolean shouldDisablePresenceReporting(boolean original) {
+        try {
+            return original || Pref.hideOnlineStatus();
+        } catch (Exception ex) {
+            Logger.printException(() -> "shouldDisablePresenceReporting failed: ", ex);
+            return original;
+        }
+    }
+
+    /**
+     * Decides whether the legacy Thrift foreground payload is published on "/t_fs".
+     *
+     * <p>False makes XplatNativeClientWrapper.setForeground receive a null payload, the path the
+     * app already takes when the UPI rollout flag is on, so the native keepalive and the
+     * isForeground bookkeeping still run.
+     *
+     * @param original the app's own decision, !DA7.A03()
+     */
+    public static boolean shouldSendLegacyPresence(boolean original) {
+        try {
+            return original && !Pref.hideOnlineStatus();
+        } catch (Exception ex) {
+            Logger.printException(() -> "shouldSendLegacyPresence failed: ", ex);
+            return original;
+        }
+    }
+}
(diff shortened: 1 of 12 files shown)


patches/src/main/kotlin/app/crimera/patches/instagram/misc/ghostMode/HideOnlineStatusPatch.kt
```diff
@@ -0,0 +1,94 @@
...
+        execute {
+
+            // DGW/UPI channel. Hooking the constructor covers all of its call sites at once.
+            PresenceWriteRequestConstructorFingerprint.method.apply {
+                val superCallIndex = indexOfFirstInstruction(Opcode.INVOKE_DIRECT)
+
+                addInstructions(
+                    superCallIndex + 1,
+                    """
+                    invoke-static {p3}, $PRESENCE_DESCRIPTOR->overridePresenceStatus($PRESENCE_STATUS_CLASS)$PRESENCE_STATUS_CLASS
```
swakwork: Dont hardcode registers unless you are inserting instructions in 0

patches/src/main/kotlin/app/crimera/patches/instagram/utils/Constants.kt
```diff
@@ -40,6 +40,11 @@ object Constants {
     const val USER_SESSION_CLASS = "Lcom/instagram/common/session/UserSession;"
     const val USER_DETAIL_VIEW_MODEL_CLASS = "Lcom/instagram/profile/fragment/UserDetailViewModel;"
     const val ORIGINAL_SOUND_DATA_INTF = "Lcom/instagram/api/schemas/OriginalSoundDataIntf;"
+    const val REALTIME_CLIENT_MANAGER_CLASS = "Lcom/instagram/realtimeclient/RealtimeClientManager;"
```
swakwork: You can moves these inside the patch itself. Constants are usual used for those who might be used on multiple patches. I don't think these will be used anywhere but this patch

patches/src/main/kotlin/app/crimera/patches/instagram/utils/Constants.kt
```diff
@@ -56,6 +61,7 @@ object Constants {
         move-result-object v%s"""
 
     const val LINKS_DESCRIPTOR = "$PATCHES_DESCRIPTOR/Links;"
+    const val PRESENCE_DESCRIPTOR = "$PATCHES_DESCRIPTOR/Presence;"
```
swakwork: Similar to the comment added on L43

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/Presence.java
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (C) 2026 piko <https://github.com/crimera/piko>
+ *
+ * See the included NOTICE file for GPLv3 §7(b) terms that apply to this code.
+*/
+
+
+package app.morphe.extension.instagram.patches;
```
swakwork: You can move this package inside `dm` and also change the class name to something like `UserPresence` so that we can quickly understand who its related to

vippium:
Is ther any way we can implement something like an option which'll show the active status for those person also, who've turned off their activity. Is it posible to implement this. @pemaismais

pemaismais:
Thanks for the review. All four are addressed, one commit per point so they are easy to check:

- 3e56d127 — the `Constants.kt` additions. You were right about the criterion: `LINKS_DESCRIPTOR` lives there because several patches use it, and mine is used by one. The constants and the three fingerprints now sit in the patch file as `private`, so `ghostMode/Fingerprint.kt` is back to being untouched.
- 44506257 — the hardcoded register. Injecting at index 0 also removed the lookup for the super call, so the block is shorter.
- 570bcf75 — `Presence` is now `dm/UserPresence`.

Leaving this as a draft until I have rebuilt it on the current `dev` base.

pemaismais:
@vippium I don't think so. When someone turns their activity status off, the server stops sending their status to you entirely, so `direct_v2/get_presence/` just comes back without them in it.

There's nothing left on the device to unhide. This patch only changes what your own client sends out, and that part runs locally, which is why it can be patched at all.

swakwork:
> * [4450625](https://github.com/crimera/piko/commit/445062574bc1717f0485ce17f95fe6a66c38aba7) — the hardcoded register. Injecting at index 0 also removed the lookup for the super call, so the block is shorter.

Do not AI everything. I said "do not hardcode variables unless you are inserting in 0 index", that doesn't mean that you need to insert at 0

vippium:
@pemaismais  can you let me know a little info also ? Like currently, if user disables the activity, then the other person will show the username, instead of Active 10 mins ago as sub heading in the chat. And the main user who disables the activity status can't see the last active status of the other users who've enabled the feature.

So, my question is :

Will it show the username or the last active counter of other users to me or not ?
Will the other person who's texting me will see the last active thing or just the username in the chat ?
Becoz I think it should show the last active counter and Active now of other users. and for other users, it should show my last active counter, rather than showing just username.

PR #1719, opened 2026-08-16
fix(instagram): adds 'igsi' to list of instagram params to sanatize

yaybrianna:


naoverse06:
This is already handled by ShareLinkSanitizer on dev branch

PR #1721, opened 2026-08-16
feat(Instagram): Add in-app profile picture viewer

ViictorNunes:
Replaces the external hand-off on "View profile picture" with an
in-app viewer, so the photo is shown inside a dialog instead of
being sent to an external app via ActivityHook.

- Add ProfilePictureViewer: a rounded floating dialog with a close
  button, a square preview of the profile picture, and
  "Open image externally" / "Download profile picture" actions
  below it, built with the app's own InstagramButton component.
- Adapt panel background and close icon color to dark/light theme
  via Utils.isDarkModeEnabled().
- Load the image on a background thread with HttpURLConnection and
  BitmapFactory, since no image-loading library is used elsewhere
  in the extension.
- Wire ProfileMoreOption's "View profile picture" option to the new
  viewer instead of ActivityHook.handleUrlIntent.
  
  

https://github.com/user-attachments/assets/267014ef-a668-4a27-a3e7-d669762bb50e

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/userprofile/ProfilePictureViewer.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/userprofile/ProfilePictureViewer.java
@@ -1,0 +1,243 @@
+/*
+ * Copyright (C) 2026 piko <https://github.com/crimera/piko>
+ *
+ * See the included NOTICE file for GPLv3 §7(b) terms that apply to this code.
+ */
+
+package app.morphe.extension.instagram.patches.userprofile;
+
+import static app.morphe.extension.instagram.utils.IgStr.str;
+
+import android.app.Dialog;
+import android.content.Context;
+import android.graphics.Bitmap;
+import android.graphics.BitmapFactory;
+import android.graphics.Color;
+import android.graphics.drawable.ColorDrawable;
+import android.graphics.drawable.ShapeDrawable;
+import android.graphics.drawable.shapes.RoundRectShape;
+import android.os.Handler;
+import android.os.Looper;
+import android.util.DisplayMetrics;
+import android.util.TypedValue;
+import android.view.Gravity;
+import android.view.View;
+import android.view.ViewGroup;
+import android.view.Window;
+import android.view.WindowManager;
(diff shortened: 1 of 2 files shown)


extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/userprofile/ProfilePictureViewer.java
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    private static class SquareFrameLayout extends FrameLayout {
+        SquareFrameLayout(Context context) {
+            super(context);
+        }
+
+        @Override
+        protected void onMeasure(int widthMeasureSpec, int heightMeasureSpec) {
+            super.onMeasure(widthMeasureSpec, widthMeasureSpec);
```
naoverse06, line 205: Respect the incoming height constraint here. in landscape or split-screen mode, forcing the image height to the full dialog width makes the panel taller than the available window, clipping the external-open and download buttons because the panel is not scrollable

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/userprofile/ProfilePictureViewer.java
```diff
@@ -0,0 +1,243 @@
...
+            dialog.setCancelable(true);
+            dialog.setCanceledOnTouchOutside(true);
+
+            LinearLayout panel = new LinearLayout(context);
+            panel.setOrientation(LinearLayout.VERTICAL);
+            panel.setLayoutParams(new ViewGroup.LayoutParams(
+                    ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.WRAP_CONTENT));
+
+            boolean isDarkMode = Utils.isDarkModeEnabled();
+            int panelColor = isDarkMode ? 0xFF121212 : 0xFFFFFFFF;
```
naoverse06, line 78: Use instagram’s theme attributes and the existing dim utilities instead of fixed light/dark colors and local dimension helpers. this keeps the dialog consistent with instagram themes and avoids duplicating existing ui primitives

naoverse06 (changes requested):

PR #1722, opened 2026-08-16
fix(Instagram): Preserve pill button contrast

naoverse06:
Related to #1648


vippium:
Are you sure ?? Bcoz after this pr, comment section return back to normal, rather than amoled. But the options menu is in amoled. Don't you think this is still inconsistent ? I think the last version was better. 

Screenshot_20260816_210412
Screenshot_20260816_210408

vippium:
IMG_20260817_220958
IMG_20260817_220833

@naoverse06 Bro I think you need to look into this. Bcoz after this PR, so many issues are there related to material theme and amoled. 

I hope you're getting it and looking into this

PR #1728, opened 2026-08-16
fix(Instagram): blank entries in all deleted messages screen

vishnuprasad0:
Follow-up to #1531

Own messages are skipped on capture (that plan existed before but was skipped later)
But the unsend hook inserted a row regardless so unsending your own message created an entry  showing as [unknown] 
It only showed in the all deleted messages screen, not in a single chat

now only marks messages that were actually captured. Entries with no content and no sender are filtered out


vippium:
Works fine for now. In future, we can improve this feature with the thumbnail thing, rather than just having entries. Which'll be much better.

vippium:
Works fine for now. In future, we can improve this feature with the thumbnail thing, rather than just having entries. Which'll be much better.

PrimePax:
@vishnuprasad0 
Bro please look into this issue, dark screen is visible even when light theme is on. 

https://github.com/user-attachments/assets/af2052a8-8227-4005-979c-b5980c4a81c7

Issue #1730, opened 2026-08-16
[Twitter] The current sensitive media bypass in 3.9.0-dev.8 doesn't work


ViniciussSantos:
So I patched 12.17.0-release.0 with 3.9.0-dev.8 (the one that includes this [fix](https://github.com/crimera/piko/pull/1723)) and the blur + ID verification are still there. I'm in Brazil, where the age gate is on by default, so this is very much daily life for me right now, and I got curious enough about why it didn't work to decompile the APK and go digging. I think I found it, and it's honestly kind of an annoying answer. The patch is fine, it's mostly hooking a pipeline the app has moved off of.

X nowadays carries two complete implementations of this whole sensitive media flow:

- **The legacy one**: LoganSquare models in `com.twitter.model.json.*` (`JsonSensitiveMediaWarning`, `JsonBlurredImageInterstitial`) plus the old View-based UI in `com.twitter.tweetview.core.*`. This is everything the patch hooks, and the mechanics are all correct, I checked. The fingerprints match, the injection points in the `$$JsonObjectMapper.parse` bridges are right, and if a tweet actually went through this path, clearing `interstitial_action` / `available_verification_options` and auto-clicking "Show" would work.

- **The new one**: timeline responses now come through GraphQL (`query HomeTimeline` requests `blurred_image_interstitial { ... interstitial_action available_verification_options }` directly), get deserialized with kotlinx.serialization into `com.x.models.interstitial.BlurImageInterstitial` (its own `InterstitialAction.AgeVerificationPrompt`, its own `AgeVerificationPromptType.Selfie/Persona`), and rendered by a Compose UI living in `com.x.sensitivemedia.*` / `com.x.ui.common.media.sensitive.*`. The timeline post media composable (`com.x.urt.items.post.media.k1`) calls into that stack. None of it is touched by the patch, so for anything rendered this way the hooks never even run.

The thing is that in 12.17.3-alpha.01 the whole `com.twitter.model.json` package is simply gone, deleted from the dex, so on newer builds these fingerprints match nothing at all.

My guess on why it looked fixed when tested: which stack you get seems to be server/feature-flag driven (both are fully wired in the release APK), and maybe the legacy interstitial still shows up for some accounts or surfaces. Either way, from an age-gated region I can confirm first hand that on 12.17.0-release.0 + 3.9.0-dev.8 nothing visibly changed.

[...]

Issue #1659, opened 2026-08-06
[Twitter/X] Sensitive posts are hidden even though show sensitive media is on


Gtoonm:
For the past few days I've seen more and more sensitive posts be hidden behind an age verification filter that says requires ID or a selfie to bypass.

Is there a way to add a patch that bypasses the age verification?

crimera:
Resolved by #1732.

Issue #1587, opened 2026-07-26
[Twitter] Show sensitive media AND age verification selfie isn't working


ax1121:
On the latest version v12.7.1-release.0 , The option in piko settings "show sensitive media" is not working despite toggling it on. 
Additionally, trying to verify my age by scanning a selfie from the internet gave me the error "something went wrong, try again later" despite multiple attempts. This is something that worked in the past. 

I downgraded to version 11.99 and I was able to verify my age by scanning a selfie from the internet. Also, the option "show sensitive media" worked fine without needing to verify my age on another device.

tyteen4a03:
Can reproduce. I've had to downgrade to 3.7.0.

reimadenx99:
Show sensitive media works for me up to 12.2.0 (piko+shim) but for most creators I still cannot see their media

LEARJS:
Same issue here. Can't verify and can't show sensitive media.

mhcp00:
"Show sensitive media" should solved with my PR #1605 when will be merged.

BlueStarBRS:
I can also not disable this mass survailance id verification bullshit and i will not do it for a f**king social Plattform but my old Version whas to old to post or commet and now i cant see 90% of my content.

Spel0:
Still happening as of the latest dev patches

crimera:
Resolved by #1732.

Issue #1106, opened 2026-05-09
[X/Twitter][Feature Request] Spoof location or similar, to avoid age verifications in places like Brazil


Kef0121:
Using a VPN did the job until recently. It doesn't work anymore.

crimera:
Resolved by #1732.

PR #1732, opened 2026-08-16
fix(Twitter): Fix `Show sensitive media` on the Compose timeline

ViniciussSantos:
Follow-up to #1723. That PR was mechanically fine, I checked the fingerprints and the injection points, and they're all correct, but it hooks to a pipeline the app has mostly moved off of, so for a bunch of users (me included) nothing changed.

More information regarding the research and solution available in https://github.com/crimera/piko/issues/1730.

Closes https://github.com/crimera/piko/issues/1730 https://github.com/crimera/piko/issues/1659, https://github.com/crimera/piko/issues/1587, https://github.com/crimera/piko/issues/1106, https://github.com/crimera/piko/issues/846, https://github.com/crimera/piko/issues/640.

### DIFF
--- a/patches/src/main/kotlin/app/crimera/patches/twitter/timeline/sensitivemediasettings/ShowSensitiveMediaPatch.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/twitter/timeline/sensitivemediasettings/ShowSensitiveMediaPatch.kt
@@ -119,9 +119,8 @@ private object SensitiveMediaInterstitialViewFingerprint : Fingerprint(
 )
 
 private object ComposeSensitiveMediaInterstitialFingerprint : Fingerprint(
-    definingClass = "Lcom/x/sensitivemedia/impl/p;",
-    name = "c",
     returnType = "Ljava/lang/Object;",
+    parameters = listOf("Landroidx/compose/runtime/Composer;", "I"),
     filters = listOf(
         methodCall(
             opcode = Opcode.INVOKE_VIRTUAL,
--- a/patches/src/main/kotlin/app/crimera/patches/twitter/timeline/sensitivemediasettings/ShowSensitiveMediaPatch.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/twitter/timeline/sensitivemediasettings/ShowSensitiveMediaPatch.kt
@@ -18,6 +18,7 @@ import app.morphe.patcher.extensions.InstructionExtensions.addInstruction
 import app.morphe.patcher.extensions.InstructionExtensions.addInstructions
 import app.morphe.patcher.methodCall
 import app.morphe.patcher.opcode
+import app.morphe.patcher.patch.PatchException
 import app.morphe.patcher.patch.bytecodePatch
 import app.morphe.patcher.util.proxy.mutableTypes.MutableMethod.Companion.toMutable
 import app.morphe.patches.all.misc.resources.ResourceType
@@ -42,6 +43,16 @@ private const val JSON_SENSITIVE_MEDIA_WARNING_CLASS_PREFIX =
     "Lcom/twitter/model/json/core/JsonSensitiveMediaWarning"
 private const val JSON_BLURRED_IMAGE_INTERSTITIAL_CLASS_PREFIX =
     "Lcom/twitter/model/json/mediavisibility/JsonBlurredImageInterstitial"
+private const val BLURRED_IMAGE_INTERSTITIAL_CLASS =
+    "Lcom/x/models/interstitial/BlurImageInterstitial;"
+
+private inline fun applyOptionalPatch(block: () -> Unit): Boolean =
+    try {
+        block()
+        true
+    } catch (_: PatchException) {
+        false
+    }
 
 internal fun getJsonFingerprint(classPrefix: String) = object : Fingerprint(
     definingClass = "$classPrefix;",
@@ -107,6 +118,22 @@ private object SensitiveMediaInterstitialViewFingerprint : Fingerprint(
     )
 )
 
+private object ComposeSensitiveMediaInterstitialFingerprint : Fingerprint(
+    definingClass = "Lcom/x/sensitivemedia/impl/p;",
+    name = "c",
+    returnType = "Ljava/lang/Object;",
+    filters = listOf(
+        methodCall(
+            opcode = Opcode.INVOKE_VIRTUAL,
+            smali = "Lcom/x/models/interstitial/MediaVisibilityResults;->getBlurImageInterstitial()$BLURRED_IMAGE_INTERSTITIAL_CLASS",
+        ),
+        opcode(
+            opcode = Opcode.MOVE_RESULT_OBJECT,
+            location = MatchAfterImmediately()
+        )
+    )
+)
+
 @Suppress("unused")
 val sensitiveMediaPatch =
     bytecodePatch(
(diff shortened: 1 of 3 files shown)


patches/src/main/kotlin/app/crimera/patches/twitter/timeline/sensitivemediasettings/ShowSensitiveMediaPatch.kt
```diff
@@ -107,6 +118,22 @@ private object SensitiveMediaInterstitialViewFingerprint : Fingerprint(
     )
 )
 
+private object ComposeSensitiveMediaInterstitialFingerprint : Fingerprint(
+    definingClass = "Lcom/x/sensitivemedia/impl/p;",
```
swakwork: Dont use the full path. The app is obfuscated so it's subjected to change

patches/src/main/kotlin/app/crimera/patches/twitter/timeline/sensitivemediasettings/ShowSensitiveMediaPatch.kt
```diff
@@ -107,6 +118,22 @@ private object SensitiveMediaInterstitialViewFingerprint : Fingerprint(
...
+    name = "c",
```
swakwork: Same as L122

patches/src/main/kotlin/app/crimera/patches/twitter/timeline/sensitivemediasettings/ShowSensitiveMediaPatch.kt
ViniciussSantos: I dropped the obfuscated names in favor of the `getBlurImageInterstitial()` call and the method signature. It still works as intended for 12.17.0-release.0 (Haven't tested on other versions though)

kitadai31:
Isn't "Compose timeline" you're referring to the New X Android (X-Lite) app?
If so, this should be done in #1567
The current piko patches does not support X-Lite, so no need to add a code for X-Lite

ViniciussSantos:
> Isn't "Compose timeline" you're referring to the New X Android (X-Lite) app? If so, this should be done in #1567 The current piko patches does not support X-Lite, so no need to add a code for X-Lite

Nope, this is the regular X app, not X-Lite. The `com.x.*` classes ship inside the regular `com.twitter.android` release builds, sitting right next to the old implementation, and which one actually renders your timeline seems to be server-driven. For me, it's the new one, everything here was developed and tested on 12.17.0-release.0, the exact version piko targets.

The two are easy to mix up. X-Lite uses the same app package (`com.twitter.android`), so apk mirrors serve its builds (12.17.3-alpha.01, 12.18.0-beta.0) as if they were alphas of the regular app. X-Lite is pure new-code, the legacy `com.twitter.model.json` package doesn't exist there at all, while regular release builds carry both stacks. I remember seeing an X employee hint at a full rewrite of the android app a while back, and X-Lite makes sense as that rewrite, so the regular app finishing the migration is probably a matter of time. Which is also why the patch tolerates the legacy classes missing instead of failing.

Since X-Lite is built from this same new codebase, these hooks might coincidentally help in #1567 too, but I haven't tested that, and it's not what this PR is about.

swakwork:
> Nope, this is the regular X app, not X-Lite. The `com.x.*` classes ship inside the regular `com.twitter.android` release builds, sitting right next to the old implementation, and which one actually renders your timeline seems to be server-driven. For me, it's the new one, everything here was developed and tested on 12.17.0-release.0, the exact version piko targets.

They have been shipping the XLite components on the primary app for a long time. His question was does this reflect it on the "New X Android" or the legacy UI. As far as we know of `com.x.*` packages are not being used on the legacy UI for most of the older components. If this patch is focused on the "New X Android" then piko patches in its current state doesn't support it.

ViniciussSantos:
> does this reflect it on the "New X Android" or the legacy UI

Both. The legacy hooks from #1723 stay, and the new hook covers the `com.x.*` stack, so whichever one the app is actually using is covered.

On scope: this isn't X-Lite support, it's the regular app being broken. On the exact build piko targets (12.17.0-release.0, regular release track, no X-Lite involved), 3.9.0-dev.8 with only the legacy hooks changed nothing for me, blur and verify-age still there. Same APK plus this PR's hook and they're gone. So the timeline the regular app serves me is rendered by the `com.x.*` code, I don't see another way to read it.

Which stack a user gets seems server/flag driven (I'm in Brazil with the age gate on aka the child safety law check, maybe that's the trigger), which would explain why #1723 looked fixed for the author and did nothing for me. And yeah, most older components are still legacy, the migration looks per-surface, the home timeline already moved.

piko already hooks `com.x.*` in the regular app btw, `handleModernShareSheetLinks` goes through `com.x.models.ContextualPost`. Dropping the new hook would leave the patch doing nothing for people on the new stack, but if you'd rather I rescope or rename anything, happy to.

vippium:
@ViniciussSantos can you let me know a little info also ? Like currently, if user disables the activity, then the other person will show the username, instead of `Active 10 mins ago` as sub heading in the chat. And the main user who disables the activity status can't see the last active status of the other users who've enabled the feature.

So, my question is : 
1. Will it show the username or the last active counter of other users to me or not ?
2. Will the other person who's texting me will see the last active thing or just the username in the chat ?
Becoz I think it should show the last active counter and Active now of other users. and for other users, it should show my last active counter, rather than showing just username.

ViniciussSantos:
@vippium I think you've got the wrong PR, this one is about the sensitive media blur on the timeline, nothing to do with DMs.

kitadai31:
@ViniciussSantos Thank you for the explanation!
I was mistaken.

I thought "Compose timeline" referred to a timeline built entirely with Compose—in other words, the New X Android app.

There is no age verification in my country, so I haven't reproduced the issue.

ViniciussSantos:
>I thought "Compose timeline" referred to a timeline built entirely with Compose—in other words, the New X Android app.

Yeah @kitadai31, I think the name I chose for the PR maybe was too confusing, should have been something in the vein of "Skip age check and display sensitive media"

swakwork:
> On scope: this isn't X-Lite support, it's the regular app being broken. On the exact build piko targets (12.17.0-release.0, regular release track, no X-Lite involved), 3.9.0-dev.8 with only the legacy hooks changed nothing for me, blur and verify-age still there. 

Its refrain from using X-Lite as is confusing because X-Lite is a completely different product. What I would like to know is are you using "New X android" aka the new look, which is part of `com.android.twitter`. Could you add a screenshot for better understanding

> piko already hooks `com.x.*` in the regular app btw, `handleModernShareSheetLinks` goes through `com.x.models.ContextualPost`. 

Because they removed legacy sharesheet started using `New X Android` sharesheet on legacy instance. But I'm not sure if it's the same with timelines aswell because that they moved, we would be having issues with Ad Blocking functionality.

ViniciussSantos:
Sorry for the delay in responding, I have been really busy with work. 

> What I would like to know is are you using "New X android" aka the new look, which is part of `com.android.twitter`. Could you add a screenshot for better understanding

Sure, here's my home timeline, For you tab:
Screenshot_20260821_020414_X
 
This is the regular app, `com.twitter.android` 12.17.0-release.0, with 3.9.0-dev.8 plus this PR. I never opted into any new look or beta that I know of, this is just what X serves my account. If this is what you mean by the new look, then yes, and regular users get served this without ever choosing anything. I honestly don't know where the line between legacy look and new look is supposed to be, so if the screenshot doesn't help, tell me what to check. 

> But I'm not sure if it's the same with timelines aswell because that they moved, we would be having issues with Ad Blocking functionality. 

"Hide ads" is on, I'm a free user (no Premium), and I haven't seen a single promoted post in a week of daily use on this build

PR #1741, opened 2026-08-17
feat(Instagram): Restore classic search recents (up to 25)

Manishrdy:
Search used to show up to 25 recent profiles on the empty-state screen. Newer builds swapped that for Meta AI suggestions plus a collapsed list you have to tap "show more" on this brings the old one back.

One thing had to happen first: the flag hook only did boolean MobileConfig overrides, and the recent count values are `long` (MobileConfig has no real int type). Extended the hook to cover long-returning flags too. Same accessor class as the boolean one already hooked, same method shape, just `J` instead of `Z`.

Tested on 439.0.0.37.89.

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/Flag.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/Flag.java
@@ -15,14 +15,21 @@ public class Flag {
 
     private String name;
     private String desc;
-    private String type;
+    private FlagType type;
     private StringSetting code;
 
     public Flag(JSONObject jsonObject) {
         try {
             this.name = jsonObject.optString("name");
             this.desc = jsonObject.optString("desc");
-            this.type = jsonObject.optString("type", "bool");
+            String rawType = jsonObject.optString("type", "bool");
+            this.type = FlagType.BOOL;
+            for (FlagType candidate : FlagType.values()) {
+                if (candidate.toString().equals(rawType)) {
+                    this.type = candidate;
+                    break;
+                }
+            }
             String codeKey = jsonObject.optString("code");
             // No override by default -- both bool and long flags use FlagState.DEFAULT
             // as the "no override" sentinel, so dev-options and override-backup restores
@@ -41,7 +48,7 @@ public class Flag {
     }
 
     public boolean isLongType() {
-        return "long".equals(type);
+        return type == FlagType.LONG;
     }
 
     public StringSetting getCode() {
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/Flag.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/Flag.java
@@ -24,11 +24,10 @@ public class Flag {
             this.desc = jsonObject.optString("desc");
             this.type = jsonObject.optString("type", "bool");
             String codeKey = jsonObject.optString("code");
-            // Long-typed flags default to an empty override (rendered as a blank
-            // numeric field) instead of the "default" sentinel bool flags use,
-            // since showing the literal word "default" in a number box reads wrong.
-            String defaultValue = isLongType() ? "" : FlagState.DEFAULT.toString();
-            this.code = new StringSetting(codeKey, defaultValue);
+            // No override by default -- both bool and long flags use FlagState.DEFAULT
+            // as the "no override" sentinel, so dev-options and override-backup restores
+            // (which don't touch this store) remain free to take effect unopposed.
+            this.code = new StringSetting(codeKey, FlagState.DEFAULT.toString());
         } catch (Exception e) {
         }
     }
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/Flag.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/Flag.java
@@ -15,14 +15,20 @@ public class Flag {
 
     private String name;
     private String desc;
+    private String type;
     private StringSetting code;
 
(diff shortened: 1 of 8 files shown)


Manishrdy:
I'll add screenshots shortly.

swakwork:
You can probably extend the support for Long Override flags and then move these flags to recommended flags as recommended flags were built to server that purpose.  Building patches for flags is kinda baseless as there are n number of flags controlling n number of things.

Manishrdy:
Reworked this as per your feedback - extended Recommended Flags to support long-typed overrides (it only handled booleans before) and moved all 8 flags from the old dedicated patch into the JSON catalog. `RestoreSearchRecentsPatch.kt` is gone; the long-flag bytecode hook in `HookFlagsPatch.kt` is unchanged from before.

Branch is force-pushed and squashed to a single commit, so this should be a clean review from scratch.

vippium:
Are you sure this patch is needed ? Bcoz we can remove those AI suggestions from classic search recents through flags itself. If you're addign it in the Recommended falgs, then it's another point.

Manishrdy:
Already this pr, actually this doesn't add a patch. Used to, but I dropped that after swakwork said the same thing on an earlier pass. The 8 flags just live in the Recommended Flags catalog now. Only real code change is extending Recommended Flags to handle numbers too, it was bool-only before, and two of these (max recent searches) are counts, not on/off toggles.

Worth calling out - even with the AI suggestions flag off, Meta's capped recent searches to the top 4 in newer builds. That count override is what actually brings back the full classic list - turning off the AI suggestions alone doesn't do it.

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/Flag.java
```diff
@@ -35,6 +41,10 @@ public String getDesc() {
         return desc;
     }
 
+    public boolean isLongType() {
```
swakwork, line 50: Instead create a FlagType enum class and handle type using enum so that we can extend support for other types as well in the future

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/Flag.java
```diff
@@ -15,14 +15,20 @@ public class Flag {
...
         try {
             this.name = jsonObject.optString("name");
             this.desc = jsonObject.optString("desc");
+            this.type = jsonObject.optString("type", "bool");
             String codeKey = jsonObject.optString("code");
-            this.code = new StringSetting(codeKey,FlagState.DEFAULT.toString());
+            // Long-typed flags default to an empty override (rendered as a blank
+            // numeric field) instead of the "default" sentinel bool flags use,
+            // since showing the literal word "default" in a number box reads wrong.
+            String defaultValue = isLongType() ? "" : FlagState.DEFAULT.toString();
```
swakwork: Let the default value be handled via FlagStates. 
Default - No override, system default
Enable - your custom default value.

We shouldn't be overriding the value by default as some might be overriding though override backups and some might override through developer options. If we hardcode the value here then they might not be able to handle via the other 2 ways

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/FlagsSharedPref.java
```diff
@@ -47,7 +47,10 @@ public static Map<String, Boolean> getAll(){
             while (keys.hasNext()) {
                 String key = keys.next();
                 String flagState = (String) flags.get(key);
-                if (flagState.equals(FlagState.DEFAULT.toString())) continue;
+                // Long-typed flags share this same pref file, so their stored
+                // value (a number, or empty) won't match either bool state --
+                // only treat it as a bool override if it actually looks like one.
+                if (!flagState.equals(FlagState.ENABLE.toString()) && !flagState.equals(FlagState.DISABLE.toString())) continue;
                 Boolean value = flagState.equals(FlagState.ENABLE.toString()) ? true:false;
                 PikoUtils.logger(key+" : "+Boolean.valueOf(value));
```
swakwork: You can comment this out. I added it for debug purpose.

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/devFlags/FlagsSharedPref.java
```diff
@@ -57,4 +60,26 @@ public static Map<String, Boolean> getAll(){
...
+        Map<String, Long> outFlags = new HashMap();
+        try {
+            JSONObject flags = INSTANCE.all();
+            Iterator<String> keys = flags.keys();
+            while (keys.hasNext()) {
+                String key = keys.next();
+                String flagState = (String) flags.get(key);
+                try {
+                    Long value = Long.parseLong(flagState);
+                    PikoUtils.logger(key+" : "+value);
```
swakwork: No need for logging here

Manishrdy:
Pushed fixes for all four.

- Flag now uses a real FlagType enum instead of comparing strings - same shape as FlagState, plain enum with lowercase toString().
- defaultValue uses FlagState.DEFAULT for both types now instead of the hardcoded "" for long flags. Turns out that brought back exactly the problem the "" was there to avoid - the numeric field would show the literal word "default"; so EditTextPref blanks that out on display now, doesn't touch what's actually persisted.
- Commented out your getAll() logger rather than deleting it, and dropped the one in getAllLong() since that was just mine mirroring yours.

naoverse06 (dismissed):

naoverse06:
Please test it on dev.8, and if there are no issues, I’ll merge it

Manishrdy:
@naoverse06 tested on dev.8; no issues found. Good to merge from my side.

PR #1745, opened 2026-08-18
feat(Instagram): Indicator for deleted messages and the ability to view content (images, videos, voice messages, etc.) externally

ViictorNunes:
**An indicator for unseen deleted messages in a conversation, as well as the ability to view the deleted content externally, as shown in the image and video below.** _I added this because I felt it was missing from Piko compared to other mods (like Honista)._

image

https://github.com/user-attachments/assets/6b7190a5-3a6b-424e-824c-22eb2c615c7f

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/UI.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/constants/UI.java
@@ -15,9 +15,12 @@ import android.graphics.Color;
 import android.graphics.PorterDuff;
 import android.graphics.PorterDuffColorFilter;
 import android.graphics.drawable.Drawable;
+import android.graphics.drawable.GradientDrawable;
 import android.util.TypedValue;
+import android.view.Gravity;
 import android.view.View;
 import android.view.ViewGroup;
+import android.widget.FrameLayout;
 import android.widget.ImageView;
 import java.util.ArrayList;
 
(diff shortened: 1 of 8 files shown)


vippium:
Id you're adding this, then try to add a thing by which we can listen the audio without external player. Means by using insta's official internal player.

ViictorNunes:
> Id you're adding this, then try to add a thing by which we can listen the audio without external player. Means by using insta's official internal player.

I can give this possibility tomorrow.

vippium:
As the PR is in progress, try to have the counter badge, rather than the dot badge. So that user can get to know exactly how many messages has been deleted from last time. And after user opened the panel, the counter badge will be gone and ready to count further deleted messages from 0.

You're getting my point what I want to say. Bcoz on dot indicator, it is not possible to view the count. And also, try to make it red, rahter than yellow. and for Material theme users, it'll follow the theme color.

ViictorNunes:
> Id you're adding this, then try to add a thing by which we can listen the audio without external player. Means by using insta's official internal player.

Unfortunately, it's not possible; I tried to find a way, but there just isn't one.

ViictorNunes:
> As the PR is in progress, try to have the counter badge, rather than the dot badge. So that user can get to know exactly how many messages has been deleted from last time. And after user opened the panel, the counter badge will be gone and ready to count further deleted messages from 0.

He stayed with the counter the way you spoke and took the opportunity for the background of the counter to follow the colors of the user's material You and be blue by default. I took advantage of and fixed a bug that the counter was displayed in chats that did not have a deleted message. 
Screenshot_20260820_192901_Instagram

Screenshot_20260820_191249_Instagram

naoverse06:
the dm header button and badge require additional ActionBar hooks, per-thread state management, custom badge ui, and theme/layout handling. since this integration has already caused spacing and compatibility issues, i don't think we should keep it, could you split the external media viewing and highest-resolution selection into a separate pr

naoverse06:
can you remove the history button and badge from the  dm header and show the count under view deleted messages in piko’s dm settings? add an is_seen column to saved messages. existing rows should be marked as seen, new rows should be saved as unseen, and opening the history should mark them all as seen. hide the count at 0 and cap it at 99+

ViictorNunes:
> can you remove the history button and badge from the dm header and show the count under view deleted messages in piko’s dm settings?

I didn't really like the idea; sometimes I have to manually check for deleted messages, and I think some other users do too. Why make it such a long process just to see deleted messages?

> add an is_seen column to saved messages. existing rows should be marked as seen, new rows should be saved as unseen, and opening the history should mark them all as seen. 

I like the idea; I'll try to apply it to the code.

> hide the count at 0 and cap it at 99+

But that feature is already in the code.

ViictorNunes:
> could you split the external media viewing and highest-resolution selection into a separate pr

I can yes

> the dm header button and badge require additional ActionBar hooks, per-thread state management, custom badge ui, and theme/layout handling. since this integration has already caused spacing and compatibility issues, i don't think we should keep it

If a problem arises, we’ll fix it. Because if that were the case, shouldn't allow the deleted message feature.

naoverse06:
this isn’t about removing the deleted messages feature. the issue is the extra actionbar hook, custom badge ui, and per-thread state needed only for the dm header. it has already caused layout issues, and “fix it when it breaks” is exactly the maintenance cost i’m trying to avoid. please remove the header button and badge. deleted messages will still be available in dm settings. keep the existing is_seen and 99+ logic, and show the unseen count under view deleted messages instead

ViictorNunes:
> the issue is the extra actionbar hook, custom badge ui, and per-thread state needed only for the dm header. it has already caused layout issues, and “fix it when it breaks” is exactly the maintenance cost i’m trying to avoid. please remove the header button and badge.

So I prefer to close the commit; that might be better.

PR #1754, opened 2026-08-19
feat(Instagram): download saved collections

Pelotrio:
## Summary

Adds a `Download collection` action to Instagram saved collection menus. It loads the collection, shows the number of posts and files, then downloads the files as one cancellable batch.

| Collection menu | Confirmation |
|:--:|:--:|
| Download collection action in the saved collection menu | Confirmation showing 270 posts and 335 files |

## How it works

Instagram loads saved collections one page at a time. The patch gets the Posts fragment for the open collection and uses Instagram's paginator until there are no pages left. It waits for each page request to finish before requesting the next one.

The Posts tab supplies the collection. The Reels tab is not added separately because its media was already present in Posts on the supported Instagram version.

Once loading finishes, the patch converts each post into download requests with the existing media parser. Each carousel item becomes a separate file, which is why the file count can be higher than the post count.

The downloader processes those requests in order and updates one notification throughout the download. A foreground service keeps the batch running when Instagram is in the background. Cancelling stops the current request and deletes its partial file.

Instagram obfuscates the paginator fields and methods. For the supported v439 target, the patcher finds them from the bytecode structure and stops with an error if it cannot identify them uniquely.

## Testing

- `./gradlew clean buildAndroid`
- `checkStringResources`: 12,842 strings checked, no issues
- GitHub Actions `Test pull request`: passed
- Patched Instagram `439.0.0.37.89` (`384510827`, arm64-v8a) with the normal Piko patch set; no patches failed
- Loaded a collection from `36` to `73` posts through the native paginator
- Cancelled at the confirmation dialog and confirmed that no downloads started
- Confirmed a `73` post collection produced `132` file requests after expanding carousels
- Downloaded a larger collection with `270` posts and `335` files
- Sent Instagram to the background during the download and confirmed that progress continued
- Cancelled from the notification and confirmed that the active download stopped
- Confirmed that the completion totals accounted for all `335` file requests


swakwork:
If you would ask me, I would say its a needless feature/patch.

dr-strange15:
I would love to have this feature if it downloads all of saved collection at once(?)

But also 1800+ lines of code for this single feature is quite a lot for future maintainance for devs😅

Pelotrio:
> If you would ask me, I would say its a needless feature/patch.
@swakwork 

For me it was one of the things that I felt like were missing but I agree this turned into quite a large change which can be hard to maintain. I tried cleaning it up a bit already and I am open for suggestions / changes how this could be improved to make this an actually useful feature that can be added. I would appreciate if we can discuss if this is worth adding so I know whether to keep working on this to make this pr clean or just keep it private for myself.

extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/download/CollectionDownloadPatch.java
```diff
@@ -0,0 +1,879 @@
...
+                }
+            });
+        } catch (Throwable error) {
+            logError("Could not add the saved collection download action", error);
+        }
+    }
+
+    private static void addNormalAction(Object actionBuilder, String title, View.OnClickListener listener)
+            throws Exception {
+        Method method = actionBuilder.getClass().getDeclaredMethod(
```
swakwork: Create an entity class for it, extend `Entity` and use it as an reflection wrapper

swakwork:
> I would appreciate if we can discuss if this is worth adding so I know whether to keep working on this to make this pr clean or just keep it private for myself.

If you are sending any additional requests to collect the collection information then we may have to halt it. Last time when someone tried to do something like that on a different project, there was a ban wave.

Pelotrio:
> > I would appreciate if we can discuss if this is worth adding so I know whether to keep working on this to make this pr clean or just keep it private for myself.
> 
> If you are sending any additional requests to collect the collection information then we may have to halt it. Last time when someone tried to do something like that on a different project, there was a ban wave.

What I do is pretty much 
`Start download collection → Instagram’s native next-page method → native cursor/state → wait until idle → request next page`

So its not much different from manually scrolling down the entire collection to have all posts loaded. I am not sure if that by itself is a problem already. Downloading itself seems scarier here in my opinion.

PR #1762, opened 2026-08-20
fix(Twitter): restore relationship actions in user lists

### LINKED ISSUE #1514: [Twitter] Follow button stops working in patched Twitter/X 12.7.0
In patched Twitter/X 12.7.0, following accounts stops working after a while. The button changes to **Following** and turns black, but the account is not actually followed. After refreshing, it returns to **Follow**.

This is not caused by X’s follow limit - I verified it three times. Rolling back to patched 12.2.0 fixes the issue.

A similar problem occurred before with Piko patches, and after it was reported, you managed to identify the cause.

RodrigoMenaM:
## Summary

- Route followers, following, subscriptions, and related user lists through the existing legacy timeline implementation instead of the XLite relationship-list path.
- Ensure both timeline action-handler factories use the active account identifier.
- Prevent list follow/unfollow actions from changing only optimistic UI state and reverting after refresh.

## Validation

- Built with .\gradlew.bat :patches:buildAndroid.
- Applied the patch to a clean X 12.7.1-release.0 APK; all 71 selected patches succeeded.
- Inspected the installed bytecode: all four relationship-list builders overwrite the follower-timeline feature-flag result with false, and both action-handler factories obtain the current user identifier.
- Manually verified that unfollowing from the Following list persists after refresh.

Fixes #1514

### DIFF
--- a/patches/src/main/kotlin/app/crimera/patches/twitter/misc/fixListRelationshipActions/FixListRelationshipActionsPatch.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/twitter/misc/fixListRelationshipActions/FixListRelationshipActionsPatch.kt
@@ -13,52 +13,39 @@ import app.morphe.patcher.extensions.InstructionExtensions.getInstruction
 import app.morphe.patcher.extensions.InstructionExtensions.instructions
 import app.morphe.patcher.patch.bytecodePatch
 import app.morphe.util.getReference
+import com.android.tools.smali.dexlib2.AccessFlags
 import com.android.tools.smali.dexlib2.Opcode
 import com.android.tools.smali.dexlib2.iface.instruction.OneRegisterInstruction
+import com.android.tools.smali.dexlib2.iface.reference.FieldReference
 import com.android.tools.smali.dexlib2.iface.reference.StringReference
 
 private const val FOLLOWER_TIMELINE_STACK_FLAG =
     "android_follower_timelines_stack_enabled"
 
-private const val TIMELINE_USER_ACTION_FACTORY =
-    "Lcom/twitter/app/common/timeline/di/view/l;"
-
-private const val TIMELINE_USER_ACTION_FACTORY_ALTERNATE =
-    "Lcom/twitter/app/common/timeline/di/view/i0;"
-
-private const val TIMELINE_USER_ACTION_HANDLER =
-    "Lcom/twitter/users/timeline/l;"
-
-private val timelineUserActionFactoryParameters =
-    listOf(
-        "Landroid/content/Context;",
-        "Landroidx/fragment/app/FragmentManager;",
-        "Lcom/twitter/safetymode/common/h;",
-        "Lcom/twitter/async/http/f;",
-        "Lcom/twitter/util/user/UserIdentifier;",
-        "Lcom/twitter/cache/twitteruser/a;",
-        "Lcom/twitter/app/common/z;",
-        "Lcom/twitter/analytics/feature/model/p1;",
-        "Landroidx/fragment/app/Fragment;",
-        "Lcom/twitter/onboarding/gating/a;",
-        "Lcom/twitter/onboarding/gating/c;",
-    )
-
 private object TimelineUserActionFactoryFingerprint : Fingerprint(
-    definingClass = TIMELINE_USER_ACTION_FACTORY,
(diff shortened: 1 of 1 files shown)


patches/src/main/kotlin/app/crimera/patches/twitter/misc/fixListRelationshipActions/FixListRelationshipActionsPatch.kt
```diff
@@ -0,0 +1,124 @@
...
+        "Landroidx/fragment/app/FragmentManager;",
+        "Lcom/twitter/safetymode/common/h;",
+        "Lcom/twitter/async/http/f;",
+        "Lcom/twitter/util/user/UserIdentifier;",
+        "Lcom/twitter/cache/twitteruser/a;",
+        "Lcom/twitter/app/common/z;",
+        "Lcom/twitter/analytics/feature/model/p1;",
+        "Landroidx/fragment/app/Fragment;",
+        "Lcom/twitter/onboarding/gating/a;",
+        "Lcom/twitter/onboarding/gating/c;",
```
crimera: please avoid hard coding these obfuscated class names. classes named as letters mean they are obfuscated they can change between updates. requiring us to have to update these names every update. A better way of doing this would be to use more stable fingeprints. sometimes this classes have strings in them you can match them in the fingerprint

RodrigoMenaM: Addressed in 410652f8. I removed the hard-coded one-letter class descriptors.

The factory fingerprint now matches on stable package prefixes, public/static method shape, stable Context/FragmentManager/UserIdentifier/Fragment parameter positions, and the UserIdentifier field write. matchAll(2..2) finds and patches both factories. The navigation fingerprint now combines the stable feature-flag string with the users/timeline package prefix.

Validation:
- .\gradlew.bat :patches:buildAndroid succeeds.
- An exclusive Morphe patch run against a clean X 12.7.1-release.0 APK applied successfully and modified the four expected classes.

swakwork:
I'm currently on Version `12.17.0-release.0` and am not having the issue, https://github.com/crimera/piko/issues/1514, may be its a base issue which got fixed

crimera:
mb, accidentally merged to main. can you redo the pull request? but this time do it on dev branch

isillo:
Unfortunately, the patches in v3.9.0-dev.10 did not fix the issue. I still cannot follow followers of other users. I open someone's profile, go to their Followers list and tap the Follow buttons, but they react instantly instead of having the tiny delay they normally have when the action works correctly. After pulling down to refresh the list, everything returns to the previous state and none of the users are actually followed. Please investigate this issue further.

naoverse06:
because it was merged into main instead of dev

isillo:
> because it was merged into main instead of dev

Is this available to the general public in Morphe App?

naoverse06:
no

isillo:
> no

Could it also be added to the dev branch and made available in Morphe within the next few hours?

kitadai31:
piko doesn't release dev patches automatically, so adding a commit to dev branch doesn't mean making it available in Morphe Manager within the next few hours

isillo:
I thought the fix had been incorporated into the main branch, so I expected it to be included in the full 3.9.0 patch release. However, following still doesn't work. What's going on?

kitadai31:
main branch has been restored to before merged this pull request because of the mistake
So the author, @RodrigoMenaM, have to open this pull request again

isillo:
@RodrigoMenaM, please open this pull request again for both dev and main asap

kitadai31:
> @ RodrigoMenaM, please open this pull request again for both dev and main asap

No, open for dev branch only

RodrigoMenaM:
Redone for dev: #1803

isillo:
patch still not included in todays v3.10.0-dev.1?

isillo:
patch still not included in 3.10.0-dev.3?! why so long wait?

isillo:
@RodrigoMenaM @kitadai31 @naoverse06 Guys, whats going on? Why hasnt the patch been available for the last three dev releases?

kitadai31:
@isillo We're moved
Read the 3 messages above, man

Btw if you really need the fix asap, disable the feature flag `android_follower_timelines_stack_enabled`

Issue #1747, opened 2026-08-18
[Twitter] "This app is out of date" error keeps showing up


MegaMinxUp:
Whenever I open the patched twitter app (latest version and applied using morphe, not the prebuilts), the error "This app is out of date" annoyingly shows up at random, making it annoying to navigate. I have uninstalled a previous version where the same error started showing up recently before doing the patch, and it still shows up on the new install (the only other change I made in morphe settings before patching is to return the old logo, nothing else)

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/ecc78857-c1d9-45ac-9da9-e62f1b730aa6" />

RodrigoMenaM:
**Correction/update:** my original report did not distinguish the PR from an earlier local compatibility backport.

That first local backport contained only the client-shutdown state guard. It was incomplete and caused a navigation regression: after switching profiles, the app could land on a root-level update screen that Back no longer escaped. That build has been withdrawn and should not be used.

The valid fix requires both parts of **Block update screen**:

- the existing full-cover UI hide/dismiss hook; and
- the shutdown-state guard added by #1763.

The PR targets `dev`, where the UI hook already exists, and now explicitly documents that its state-guard commit must not be backported alone. A combined compatibility build was applied and checked at runtime: cold launch and profile switching in both directions stayed on the normal app UI, with no update-only activity or fatal signal. The final reviewer-requested fingerprint/utility refactor also builds successfully.

Issue #1717, opened 2026-08-16
A lot of app shortcuts were created


LinUwUxCat:
I installed piko through morphe manager as instructed in the readme. Imported my settings from the previous installation i had, which was 2.0.0-dev.11, and looked around in the settings after. The first thing i did was change the app icon, after clicking on one the app crashed, i restarted it, and after that changing icon worked fine, except it wouldnt actually move the radio button to the newly selected icon, but it did actually change, so whatever. Then i went to disable the native downloader (i did not see the inline button option below which was what i actually wanted) and doing that crashed the app again. When i went back to open the app, the apps list looked like this : 

<img width="1215" height="1956" alt="Image" src="https://github.com/user-attachments/assets/3964cf8b-5b70-4d8b-8e2e-e08cddbd0494" />

The one i set through the settings is at the very top right.

This can likely be fixed with a reinstall, and i think this is pretty funny, but i figured a report is the best thing to do in case this can be fixed

sudoshindo:
I've experienced this problem.

The change app icon feature breaks Twitter, it doesn't work properly.

kitadai31:
This should already be fixed in [v3.9.0-dev.3](https://github.com/crimera/piko/releases/tag/v3.9.0-dev.3)

sudoshindo:
> This should already be fixed in [v3.9.0-dev.3](https://github.com/crimera/piko/releases/tag/v3.9.0-dev.3)

Thanks for letting us know. Hopefully 3.9.0 is soon. I keep getting the "This app is out of date" screen.

PR #1763, opened 2026-08-21
fix(Twitter): address client shutdown behind update screen

RodrigoMenaM:
## Scope

This PR extends the existing **Block update screen** patch. The fix is the combination of two protections:

1. The existing full-cover UI hook hides and dismisses the update screen.
2. This PR makes the underlying client-shutdown state report `false` while the same setting is enabled.

The state guard must not be shipped as a standalone backport without the full-cover UI hook.

## Why both hooks are needed

X persists `is_shutdown` after its remote minimum-version check. That state is consumed outside the dialog path by launcher-entry handling and the client-shutdown network interceptor. Handling only the visible dialog can therefore leave the shutdown state active.

Conversely, an earlier local compatibility build contained only the state guard. That incomplete backport regressed account-switch navigation: switching profiles could land on a root-level update screen that Back no longer escaped. That build has been withdrawn. The corrected compatibility build combines the existing UI hook with the state guard.

The `dev` base of this PR already contains the full-cover UI hook; this diff adds the complementary state guard.

## Change

- Identify the shutdown-state class by its stable `is_shutdown` and `shutdown_min_version` preference keys.
- Fingerprint its `isShutdown()` method directly within that resolved class.
- When **Block update screen** is enabled, return `false` to client-shutdown consumers.
- When disabled, continue through X's original stored-state path.
- Keep the visible full-cover fallback tied to the same setting.
- Place the shared setting lookup in `Utils`, following maintainer review.

## Verification

- Final revision: `:patches:buildAndroid` succeeds.
- The combined implementation was applied to a compatible local X package and both injected hooks were confirmed in the installed DEX.
- Cold launch stayed on the normal main activity.
- Profile switching was exercised in both directions without the update screen, update-only activity, or a fatal runtime signal.
- The reviewer-requested organization/fingerprint refactor does not change the injected runtime branch; its final revision is build-verified.

Addresses #1747 and the continued report in #1717.

### DIFF
--- a/extensions/twitter/src/main/java/app/morphe/extension/twitter/Pref.java
+++ b/extensions/twitter/src/main/java/app/morphe/extension/twitter/Pref.java
@@ -144,8 +144,12 @@ public class Pref {
 
     public static boolean redirect(TabLayout$g g) {return Utils.redirect(g);}
 
+    public static boolean blockUpdateScreen() {
+        return Utils.getBooleanPref(Settings.MISC_BLOCK_UPDATE_SCREEN);
+    }
+
     public static void blockUpdateScreen(View view) {
-        if (Utils.getBooleanPref(Settings.MISC_BLOCK_UPDATE_SCREEN) && view != null) {
+        if (blockUpdateScreen() && view != null) {
             if (view.getParent() instanceof ViewGroup container &&
                     container.getParent() instanceof ViewGroup scrollView) {
                 // Hide the alert dialog container first
--- a/extensions/twitter/src/main/java/app/morphe/extension/twitter/Pref.java
+++ b/extensions/twitter/src/main/java/app/morphe/extension/twitter/Pref.java
@@ -144,12 +144,8 @@ public class Pref {
 
     public static boolean redirect(TabLayout$g g) {return Utils.redirect(g);}
 
-    public static boolean blockUpdateScreen() {
-        return Utils.getBooleanPref(Settings.MISC_BLOCK_UPDATE_SCREEN);
-    }
-
     public static void blockUpdateScreen(View view) {
-        if (blockUpdateScreen() && view != null) {
+        if (Utils.blockUpdateScreen() && view != null) {
             if (view.getParent() instanceof ViewGroup container &&
                     container.getParent() instanceof ViewGroup scrollView) {
                 // Hide the alert dialog container first
--- a/patches/src/main/kotlin/app/crimera/patches/twitter/misc/blockUpdateScreen/BlockUpdateScreenPatch.kt
+++ b/patches/src/main/kotlin/app/crimera/patches/twitter/misc/blockUpdateScreen/BlockUpdateScreenPatch.kt
@@ -9,6 +9,7 @@ package app.crimera.patches.twitter.misc.blockUpdateScreen
 import app.crimera.patches.twitter.misc.settings.settingsPatch
 import app.crimera.patches.twitter.utils.Constants.COMPATIBILITY_X
 import app.crimera.patches.twitter.utils.Constants.PREF_DESCRIPTOR
+import app.crimera.patches.twitter.utils.Constants.UTILS_DESCRIPTOR
 import app.crimera.patches.twitter.utils.enableSettings
 import app.morphe.patcher.Fingerprint
 import app.morphe.patcher.InstructionLocation.MatchAfterWithin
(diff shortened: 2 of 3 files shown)


extensions/twitter/src/main/java/app/morphe/extension/twitter/Pref.java
```diff
@@ -144,8 +144,12 @@ public static boolean shareImageAutoCleanup() {
 
     public static boolean redirect(TabLayout$g g) {return Utils.redirect(g);}
 
+    public static boolean blockUpdateScreen() {
+        return Utils.getBooleanPref(Settings.MISC_BLOCK_UPDATE_SCREEN);
+    }
+
     public static void blockUpdateScreen(View view) {
```
swakwork, line 147: Move this function to utils

patches/src/main/kotlin/app/crimera/patches/twitter/misc/blockUpdateScreen/BlockUpdateScreenPatch.kt
```diff
@@ -46,6 +49,15 @@ private object FullCoverDialogInflateFingerprint : Fingerprint(
     )
 )
 
+private object ClientShutdownStateFingerprint : Fingerprint(
+    name = "<init>",
+    returnType = "V",
+    strings = listOf("is_shutdown", "shutdown_min_version"),
+    custom = { _, classDef ->
+        classDef.type.startsWith("Lcom/twitter/subsystem/clientshutdown/")
```
swakwork: You can use `definingClass`

patches/src/main/kotlin/app/crimera/patches/twitter/misc/blockUpdateScreen/BlockUpdateScreenPatch.kt
```diff
@@ -78,6 +90,26 @@ val blockUpdateScreenPatch =
         )
 
         execute {
+            val isShutdownMethod =
+                ClientShutdownStateFingerprint.classDef.methods.single { method ->
```
swakwork: Any reason why you have not created fingerprint for it

RodrigoMenaM:
Correction to my earlier local validation: the first compatibility build I tested was state-only, not the full combined patch. It suppressed the persisted shutdown flag but omitted the existing full-cover UI hook. That incomplete backport regressed account-switch navigation and has been withdrawn.

The working implementation uses both hooks. I rebuilt and installed that combination, confirmed both injections in the installed DEX, and tested cold launch plus profile switching in both directions without the update wall or an update-only activity. I also updated the PR description so nobody backports the state guard alone.

The latest commit addresses all three review notes: the setting helper is in `Utils`, the class fingerprint uses `definingClass`, and `isShutdown()` has its own fingerprint scoped to the preference-key-bearing class. The final revision passes `:patches:buildAndroid`.

extensions/twitter/src/main/java/app/morphe/extension/twitter/Pref.java
RodrigoMenaM: System.Collections.Hashtable.Body

patches/src/main/kotlin/app/crimera/patches/twitter/misc/blockUpdateScreen/BlockUpdateScreenPatch.kt
RodrigoMenaM: System.Collections.Hashtable.Body

patches/src/main/kotlin/app/crimera/patches/twitter/misc/blockUpdateScreen/BlockUpdateScreenPatch.kt
RodrigoMenaM: System.Collections.Hashtable.Body

extensions/twitter/src/main/java/app/morphe/extension/twitter/Pref.java
swakwork: ?

extensions/twitter/src/main/java/app/morphe/extension/twitter/Pref.java
```diff
@@ -144,12 +144,8 @@ public static boolean shareImageAutoCleanup() {
 
     public static boolean redirect(TabLayout$g g) {return Utils.redirect(g);}
 
-    public static boolean blockUpdateScreen() {
```
swakwork, line 147: Yo keep this in Pref. I was taking about the `blockUpdateScreen(View view)` 🤦

swakwork:
> Correction to my earlier local validation: the first compatibility build I tested was state-only, not the full combined patch. It suppressed the persisted shutdown flag but omitted the existing full-cover UI hook. That incomplete backport regressed account-switch navigation and has been withdrawn.
> 
> The working implementation uses both hooks. I rebuilt and installed that combination, confirmed both injections in the installed DEX, and tested cold launch plus profile switching in both directions without the update wall or an update-only activity. I also updated the PR description so nobody backports the state guard alone.
> 
> The latest commit addresses all three review notes: the setting helper is in `Utils`, the class fingerprint uses `definingClass`, and `isShutdown()` has its own fingerprint scoped to the preference-key-bearing class. The final revision passes `:patches:buildAndroid`.

Hopefully next time I hear back from a human not an AI

PR #1767, opened 2026-08-21
docs(Instagram): State image viewing side effects in description

naoverse06:
State that the "Improve image viewing" option may reduce performance or cause instagram to switch to an older layout


kitadai31:
Can you also include information regarding the bloating of the app data (not cache)?

kitadai31:
thx

PR #1771, opened 2026-08-24
feat(Instagram): Integrate `Disable onboarding permission prompts` patch into `Disable analytics`

kitadai31:
The "Disable onboarding permission prompts" patch is only useful when "Disable analytics" is applied, so these two patches should be merged.

Also removed the old experiment-flag and legacy consent-URI approaches code, which no longer worked for any users on the current supported Instagram version.


extensions/instagram/src/main/java/app/morphe/extension/instagram/patches/Links.java
```diff
@@ -122,9 +120,6 @@ public static void interceptUri(URI uri) throws IOException{
                         || host.contains("graph.facebook.com")
                         || path.contains("/logging_client_events")) {
                     shouldBlockUri = DISABLE_ANALYTICS;
-                } else if (path.contains("/consent/existing_user_flow/")
```
swakwork, line 125: We do need to block these uri right?

patches/src/main/kotlin/app/crimera/patches/instagram/links/misc/DisableAnalyticsPatch.kt
```diff
@@ -23,5 +47,41 @@ val disableAnalyticsPatch =
...
             enableSettings("disableAnalytics")
+
+            // Disable onboarding permission prompts.
+            // If this patch is applied and the setting is on, the "Set up on new device" screen,
+            // which requests permission for location and contacts, keeps reappearing even after
+            // you have allowed or skipped it.
+            // This is caused by information that the prompt has already been seen is not sent to
+            // the server.
+            // Therefore, this patch needs to block it as well.
+            IgBloksFullScreenOpenFingerprint.apply {
```
swakwork, line 58: The simpler way would have been including the patch in `dependsOn` so that we don't loose the original hash/commit. Anyways...

kitadai31: Oh... 🤦‍♂️
Is it fine?

FYI, there was an issue where the patch was missing from Piko settings > About > Patch information, and it was easier to remove this patch than to fix that issue (Fixing this includes an addition of a string and localization for every languages)

naoverse06:
lgtm

PR #1774, opened 2026-08-25
docs: Improve README.md

kitadai31:
* Organize HTML tags
* Change "Morphe Manager" and "Morphe CLI" headers from simple bold text to actual headers
* Do not hide the Morphe Manager steps inside an details tag
  * This is the most important and in-demand part of this README, so there is no need to hide it
* Make the "Add Piko to Morphe" deep link bold
* Hide the X-Shim related information into "Open old Information" details tag
  * This makes Readme cleaner, and reduces the burden on the reader
  * While the entire X-Shim section can be removed in the future, the upcoming v3.9.0 stable release will be the first version without X-Shim for general users, so it is best to keep it inside the details tag for the time being

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,16 +1,14 @@
 <div align="center">
 
-<p align="center">
-    <img src="https://socialify.git.ci/crimera/piko/image?forks=1&language=1&name=1&owner=1&pattern=Circuit%20Board&stargazers=1&theme=Auto" alt="piko" width="640" height="320" />
-</p>
+<img src="https://socialify.git.ci/crimera/piko/image?forks=1&language=1&name=1&owner=1&pattern=Circuit%20Board&stargazers=1&theme=Auto" alt="piko" width="640" height="320" />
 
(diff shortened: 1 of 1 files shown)


kitadai31:
By the way, the "✨ Stargazers over time" line chart has been broken since July.
The graph image only says `failed to talk with github api: {"message":"Requires authentication","documentation_url":"https://docs.github.com/rest/activity/starring#list-stargazers","status":"401"}`

This is explained in this article:
https://www.star-history.com/blog/github-stargazer-api-restriction/

To fix this, you can move from `starchart.cc` to `star-history.com`.
This site supports this new restriction.
This method requires @crimera (repo owner) to issue a PAT token for stargazer API and generate an image URL.

Alternatively, since that article states the following, simply waiting might be also an option.

> *"**Update**: The GitHub team is aware of the impact and is actively working on a solution."*

Or, just commenting out or removing the "Stargazers over time" for now is also an option.

PR #1781, opened 2026-08-26
improve(Instagram): Refine anonymous DM handling

naoverse06:
Refines anonymous DM settings and manual read receipt handling. Disable typing status and View direct messages anonymously now appear only in the Ghost section instead of being duplicated in both the Ghost and DM settings. The temporary allowance used by Mark as read is now limited to the thread performing the action, it does not affect concurrent read operations. The updated descriptions explain how anonymous viewing works with manual read receipts and Instagram's native Show read receipts setting

## Testing
- `.\gradlew.bat test :patches:checkStringResources --no-daemon --console=plain`
- Tested on Instagram v439.0.0.37.89 with Piko v3.9.0-dev.9

### DIFF
--- a/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/ScreenBuilder.java
+++ b/extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/ScreenBuilder.java
@@ -198,26 +198,6 @@ public class ScreenBuilder {
     public void dmSection() {
         if (!(SettingsStatus.dmSection())) return;
 
-        if (SettingsStatus.disableTypingStatus) {
-            addPreference(
-                    helper.switchPreference(
-                            str("piko_disable_typing_status"),
-                            "",
-                            Settings.DISABLE_TYPING_STATUS
-                    )
-            );
-        }
-
-        if (SettingsStatus.viewDmAnonymously) {
-            addPreference(
-                    helper.switchPreference(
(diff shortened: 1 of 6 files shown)


ar5to:
Could we make the Ghost icon features more customizable? For example, instead of enabling everything at once, it would be useful to have individual toggles for things like View DMs anonymously, Hide Story views, etc.

It’s a bit hard to explain, but if you try InstaEclipse, you’ll probably understand what I mean. Their approach makes it much clearer which Ghost features you want to enable or disable individually.

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/ScreenBuilder.java
```diff
@@ -198,26 +198,6 @@ public void buildDeveloperSection() {
     public void dmSection() {
         if (!(SettingsStatus.dmSection())) return;
 
-        if (SettingsStatus.disableTypingStatus) {
-            addPreference(
-                    helper.switchPreference(
-                            str("piko_disable_typing_status"),
-                            "",
-                            Settings.DISABLE_TYPING_STATUS
```
swakwork, line 206: Any reason to remove this feature ?

naoverse06: same setting should not be shown twice, i only removed the duplicate entry from dm

PR #1678, opened 2026-08-11
fix(Instagram): Prevent notification registration crash

naoverse06:
Fixes a startup crash on Android 12 and later when instagram registers or refreshes its push notification token. instagram passes 0 to the flags argument of PendingIntent.getBroadcast(), but Android requires either FLAG_IMMUTABLE or FLAG_MUTABLE. adds the Fix notification crash compatibility patch, preserving the original request code and Intent while passing FLAG_IMMUTABLE through a separate register. the target method, API signature, register layout, and original zero value are validated before modifying the bytecode

## Testing
- `.\gradlew.bat :patches:check --rerun-tasks --console=plain`
- `.\gradlew.bat :patches:clean :patches:buildAndroid --console=plain`
- Tested on Instagram 439.0.0.37.89 with Piko 3.9.0-dev.7
- Confirmed the startup crash is fixed on an affected Samsung device running Android 16

Closes #1425

PR #1803, opened 2026-09-01
fix(Twitter): restore relationship actions in user lists

### LINKED ISSUE #1514: [Twitter] Follow button stops working in patched Twitter/X 12.7.0

RodrigoMenaM:
Recreates #1762 against `dev` after the accidental merge to `main` was reverted.

## Summary

- Route followers, following, subscriptions, and related user lists through the existing legacy timeline implementation instead of the XLite relationship-list path.
- Ensure both timeline action-handler factories use the active account identifier.
- Prevent list follow/unfollow actions from changing only optimistic UI state and reverting after refresh.

## Validation

- Reapplied the original two commits cleanly onto the current `dev` branch.
- `git diff --check` passes.
- The original implementation built successfully and was manually verified on X 12.7.1; unfollowing from the Following list persisted after refresh.

Fixes #1514


naoverse06:
Should this follow #1678 by removing name and default then adding the patch as a dependency of settingsPatch?

naoverse06:
did you test it on 12.19.1?

naoverse06:
does it need to be made into a separate patch?

patches/src/main/kotlin/app/crimera/patches/twitter/misc/fixListRelationshipActions/FixListRelationshipActionsPatch.kt
```diff
@@ -0,0 +1,108 @@
...
+            parameters[8].type == "Landroidx/fragment/app/Fragment;" &&
+            methodDef.implementation?.instructions?.any { instruction ->
+                instruction.getReference<FieldReference>()?.type ==
+                    "Lcom/twitter/util/user/UserIdentifier;"
+            } == true
+    },
+)
+
+private object UserTimelineNavigationFingerprint : Fingerprint(
+    strings = listOf(FOLLOWER_TIMELINE_STACK_FLAG),
```
naoverse06: Use `matchAll(expectedRange)` and verify the final method count and signatures, throw `PatchException` if they do not match

patches/src/main/kotlin/app/crimera/patches/twitter/misc/fixListRelationshipActions/FixListRelationshipActionsPatch.kt
```diff
@@ -0,0 +1,108 @@
...
+                        }.location.index
+                val resultIndex =
+                    method.instructions
+                        .drop(flagIndex + 1)
+                        .first { it.opcode == Opcode.MOVE_RESULT }
+                        .location.index
+                val resultRegister =
+                    method.getInstruction<OneRegisterInstruction>(resultIndex).registerA
+
+                method.addInstructions(resultIndex + 1, "const/4 v$resultRegister, 0x0")
```
naoverse06: Match the flag lookup and its following boolean `MOVE_RESULT`. read the register through `registersUsed` and throw `PatchException` if the match is missing or ambiguous

kitadai31:
I tested this PR.
This patch restores the legacy (Non-XLite, Non-Jetpack-Compose) follow/follower list screen, right?
This makes sense because the legacy version doesn't have the problem.

This can be achieved more easily.
You can use the existing `FeatureSwitchPatch.java` to disable the `android_follower_timelines_stack_enabled` feature flag from a patch.
In fact, I confirmed that simply disabling this feature flag from the Piko settings restores the legacy follower list screen and allows follow/unfollow actions to work correctly. (Using piko v3.9.0 stable)
Only disabling the `android_follower_timelines_stack_enabled` flag is needed to fix the problem.

Also, I think the patch name or description should mention that this patch restores the old follow/follower list screen.
Simply saying "fix" doesn't convey the effect of this patch—that it restores the old screen—and users might wonder why piko has the old screen even though they haven't disabled the flag.

crimera:
@RodrigoMenaM if you could do the suggestions from kitadai that'd be great. but if you don't have the time, I can do it for you but I'll set you as the author so you get the credit, you camd up with the patch after all.

RodrigoMenaM:
Thanks @crimera, appreciate it! Feel free to apply the simplification from kitadai31 and name it Restore legacy follower lists. My Pixel is still on 12.7.1, so I have not tested 12.19.1.

PR #1822, opened 2026-09-02
fix(Instagram): Recover piko settings after crash the app

naoverse06:
Instagram can initialize piko settings before the shared android context becomes available while recovering from a crash. BaseSharedPref then keeps an unavailable settings store for the rest of the process and returns default values, causing the "welcome to piko" dialog to appear and saved piko settings to look reset. resolve the settings store when the context becomes available. existing values remain intact, and reads continue to return their defaults while no context exists


vippium:
And what about overrides ??

PR #1835, opened 2026-09-05
feat(instagram): reflect effective toggle state in About patch list

jhult:
## Summary

The About patch list previously reported only whether each patch was **compiled in**, which never changes after install. This branch makes it reflect the **effective** state: the patch applied AND its toggles are currently on.

## Changes

- **Effective status:** About entries now report `flag AND toggle(s)` so a user's current preference is reflected instead of static compile-in status.
- **Live refresh:** the status map is re-derived every time the About screen is opened, so toggle changes appear without restarting.
- **`effective()` helper:** centralizes the flag/toggle AND logic, replacing hand-written `flag && SharedPref.getBooleanPref(...)` expressions.
- **Refactor:** `aboutSection()` now builds its own status map via   `loadStatusMap()` instead of receiving and mutating the shared `FLAGS` map from the caller (removed the now-unused `TreeMap` parameter).
- **Fixes:** `loadStatusMap()` builds a fresh map, so keys dropped from the patch set no longer linger in the About list.

## Verification

- `:extensions:instagram:compileDebugJavaWithJavac` — passes
- `:extensions:instagram:assembleDebug` — passes
- Unit tests: module has no unit tests (`NO-SOURCE`)

> Note: AI assisted in creating this fix and this pull request.


naoverse06:
i don't think this is necessary

PR #1840, opened 2026-09-06
feat(instagram): add long press inbox icon to toggle ghost mode

s1ddhants1:


swakwork:
I personally am against long presses as they are not a good UI/UX design.

If you are added it,  don't add it as a separate patch. Add it as a part of ghost mode patches.

s1ddhants1:
@swakwork yet it is a way for users to toggle ghost mode without heading to settings every time they need to toggle it and it doesn't mess with the native instagram ui too much unlike the action bar customization ghost mode icons

I don't understand what do you mean by "as a part of ghost mode patches" ? which ghost mode patch are you referring to? they're all split into patches for viewing DMs, live and stories anonymously separately

s1ddhants1:
okay shit I understand if it's patched without any of the ghost mode patches it will cause issues, 
let me implement it like the action bar utilities listed in patch dependencies

s1ddhants1:
yo @swakwork how about now?

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/ScreenBuilder.java
```diff
@@ -251,6 +251,14 @@ public void ghostSection() {
                 )
         );
 
+        addPreference(
```
swakwork, line 254: You need to add a condition if any of the ghost mode patch has been included.

swakwork:
> @swakwork yet it is a way for users to toggle ghost mode without heading to settings every time they need to toggle it.

Though long presses are easier, it doesn't have any visual cue that it exists. Thats why I haf added a button for ghost mode that user can set in action bar, for ease of use.

s1ddhants1:
@swakwork the user can find the toggle and it is undeniable that it improves qol

extensions/instagram/src/main/java/app/morphe/extension/instagram/settings/preference/ScreenBuilder.java
s1ddhants1: > You need to add a condition if any of the ghost mode patch has been included.

ghostSection already has an early return at the very top of the method why add bloat

swakwork:
Open mount support as a new PR please

s1ddhants1:
Done! Opened mount support as a separate PR in #1861 and cleaned up this PR to only contain the long press ghost mode toggle.

Issue #1379, opened 2026-06-19
[X bug] Disable auto timeline scroll on launch bug


heydubey:
Disable auto timeline scroll on launch, probably means timeline will refresh but not automatically scroll to the latest posts. 

This isn't working for me, I have the settings off, but it still won't autoscroll up to the latest post, i have to manually scroll up. 

When I turn it off, shouldn't it automatically go to the latest post.

TLDR: Disable auto timeline scroll on launch is turned off, and it still won't autoscroll to latest post.

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/94db4971-7a87-486f-a81a-c48695b0edaa" />

Issue #1363, opened 2026-06-18
[Twitter, BUG] Timeline still auto-refreshes on startup (Twitter v11.95.0+)


xiaoyaoyaowww:
The "Disable auto timeline scroll on launch" patch is no longer working on Twitter v11.95.0+. The timeline still auto-refreshes on startup.

MtndaleRedneck:
Hasn't worked for me since v11.20. However v20 has other issues. The DMs and notifications tabs don't work.

MtndaleRedneck:
@swakwork @crimera I've been reporting this same issue for months. Many people on the telegram chat reporting the same.

heydubey:
@xiaoyaoyaowww Disable auto timeline scroll on launch, probably means timeline will refresh but not automatically scroll to the latest posts. 

This isn't working for me, I have the settings off, but it still won't autoscroll up to the latest post, i have to manually scroll up

xiaoyaoyaowww:
> @xiaoyaoyaowww Disable auto timeline scroll on launch, probably means timeline will refresh but not automatically scroll to the latest posts. 
> 
> This isn't working for me, I have the settings off, but it still won't autoscroll up to the latest post, i have to manually scroll up

No, now that you've enabled this option, the next time you open Twitter, it will automatically refresh your feed and take you back to the top.

And the previous behavior was: enabling this option prevented the feed from auto-refreshing and kept your timeline position unchanged.

heydubey:
> No, now that you've enabled this option, the next time you open Twitter, it will automatically refresh your feed and take you back to the top.
> 
That's exactly what I want to happen, but it doesn't matter of I turn this option on or off, it refreshes regardless. It refreshes my feed but never takes me to the latest post. It will stay on the post i was reading last. I want it to autoscroll to the latest post after the refresh.

MtndaleRedneck:
You have the opposite problem from us. That's not how the stock twitter app works. It's supposed to hold your spot as far as I'm aware but it sometimes failed which is what this patch is for. I'm having now that every time I open the app it scrolls up regardless of if I have the setting on or off. Regardless if you want it to refresh at any time hit the X/Twitter icon at the top of the page or the Home button on the bottom to take you to the top.

heydubey:
> Regardless if you want it to refresh at any time hit the X/Twitter icon at the top of the page or the Home button on the bottom to take you to the top.

Thanks, that helped!

Qliphort8:
[...]

PR #1844, opened 2026-09-07
fix(Twitter): Restore timeline position on startup

### LINKED ISSUE #778: Timeline scroll
I have 'auto timeline scroll on launch' disabled. It still refreshes to the top of my feed whenever I exit and re-enter the app. It should be keeping my place where I'm up to in the timeline.

naoverse06:
Fix "Disable auto timeline scroll on launch" losing the reading position in the For You. capture the visible posts and their offsets, preserve the saved position during startup loading, and retry restoration when the matching post becomes available. also respect the setting when it is turned off. the previous implementation unconditionally overrode the home cold-start decision, so disabling the setting did not restore Twitter's original behavior

Closes #778, #1363, #1379

### DIFF
--- a/extensions/twitter/src/main/java/app/morphe/extension/twitter/Pref.java
+++ b/extensions/twitter/src/main/java/app/morphe/extension/twitter/Pref.java
@@ -69,6 +69,11 @@ public class Pref {
     public static boolean showSourceLabel() {
         return Utils.getBooleanPref(Settings.TIMELINE_SHOW_SOURCE_LABEL);
     }
+
+    public static boolean disableAutoTimelineScroll() {
+        return Utils.getBooleanPref(Settings.TIMELINE_DISABLE_AUTO_SCROLL);
+    }
+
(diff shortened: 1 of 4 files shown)


naoverse06:
[patches-3.10.0-dev.5.zip](https://github.com/user-attachments/files/31918200/patches-3.10.0-dev.5.zip)

swakwork:
Is it still worthy to put effort on the legacy version? Since 12.20, the have moved to "newX" builds which I don't think has legacy code base.

naoverse06:
yes i don't have much left planned for the legacy version, i only have one thing left, but once that's finished, i plan to focus on instagram

stradus64:
does this only apply to For You or also to home/Following? (on latest stable I sometimes still lose position in Following)

abeoto180:
I also confirmed that this only works in the For You tab (v3.10.0-dev.6). I hope this will also work in the Following tab (latest).

PR #1863, opened 2026-09-10
feat(Instagram): Reorder Navigation Tabs

ViictorNunes:
Added a feature to reorder Instagram tabs however the user likes. I renamed the patch and its description, and renamed the tab in the settings, as shown in the video. I also added two new hide buttons.

NOTE: I haven't yet found a way to reorder or hide the `news ` and `share `buttons, so I've disabled them in the reordering window.

https://github.com/user-attachments/assets/a7a74dfc-ece0-460a-be7f-4c0c12062764


naoverse06:
#1867

PR #1865, opened 2026-09-11
fix(newx): refresh individual media items after publication

WayneShao:
## Problem

On OEM system clone profiles, X can run as the same package name under Android user 999 while the DownloadManager provider is installed only for user 0. Public media downloads from the clone can therefore fail at the cross-user provider/storage boundary even when the owner profile works.

## Change

After a successful MediaStore publication, request a single-file media scan for the provider-allocated path and notify the exact published URI. The scan is best-effort and isolated from download rollback; it never scans a volume or rewrites the owning user's path.

This is a small prerequisite PR for #1866. It targets `x-lite` and has been rebased onto upstream commit `68170cd326aeb1c3c5e39b3fc66468dda5782d73`.

## Validation

- Combined latest-source build with the clone-download branch: X `12.25.2-prod.01` / versionCode `312252001`; all 39 selected patches applied, no failed patches.
- 209 extension tests and 40 patch tests passed; NewX resolver lint passed.
- Added application-link audit: 3 introduced app references checked, 0 unresolved.
- The resulting package was signed with the existing local test certificate and installed with `adb install -r --no-incremental` on Xiaomi HyperOS 4 / Android 17 and OnePlus ColorOS 16 / Android 16, retaining user data.
- Scanner callback behavior was previously observed on both clone profiles in the combined implementation. Immediate gallery redraw remains vendor/UI controlled.

The standalone helper does not include the clone transport. A dedicated Android integration test for scanner/query failures remains a follow-up item.


crimera:
Hello, i decided to refactor the whole media saving thing, newx now uses saf and an in house downloader.  do this changes still apply?

WayneShao:
Thanks for the refactor. After updating, my current assessment is that the new SAF + in-app downloader addresses the practical problem that motivated this PR and #1866. I am comfortable with these PRs being closed as superseded, but I want to explain the original problem, our approach, and the remaining storage difference accurately.

### Original behavior and what we were trying to fix

On my Xiaomi (HyperOS 4 / Android 17) and OnePlus 13T (ColorOS 16 / Android 16), the OEM Dual Apps / App Cloner instance runs as Android user `999`, while the main X instance runs as user `0`.

The main instance could download normally, but the clone encountered download-start/permission failures in the old DownloadManager-based flow. We observed that the system DownloadProvider was installed for user 0, not separately for user 999, and the failures were consistent with a cross-user download/storage boundary. That package-installation fact alone does not prove the exact routing or cause of every individual failure.

When downloads did succeed in the earlier clone-storage setup, including with our compatibility implementation, the files were saved under `/storage/emulated/999/...`. In my setup they were visible through the OEM gallery/file-manager integration with clone storage, rather than behaving like ordinary files saved in the owner's storage. We also encountered a separate visibility problem: a file could exist and the download could report success before the gallery reflected the new media.

The goal was reliable public image/video downloads from the clone, correct publication into its storage/media library, and timely media discovery, while keeping the main instance's behavior unchanged.

### Our proposed approach

- #1866 added an opt-in compatibility route for affected clone profiles: transfer the public media in the X app process, validate the response, and publish through the current user's MediaStore. It also handled partial-download cleanup/name conflicts and preserved the image-size fallback for unavailable `name=orig` variants.
- #1865 supplied the small publication follow-up: request a single-file media scan and notify the published URI, so a completed file would be discoverable without a whole-storage scan. This was intended to help indexing; it did not guarantee an immediate redraw of every OEM gallery.

### What changed with the upstream implementation

I installed X `12.28.0-prod.01`, patched from `x-lite` at `0e30ed414f459dbbd3fb73227936d5db89662908`, on both phones **without either local PR**. My current usage assessment is that the new flow resolves the practical download problem.

The new inline-download implementation transfers in the app and writes through a user-selected SAF document URI. It removes the old DownloadManager/staging/publication bridge from that flow, so the particular compatibility route we added is no longer needed for this use case.

Read-only ADB checks confirmed that both clone instances now have their image and video destinations set to:

```text
content://0@com.android.externalstorage.documents/tree/primary%3ADownload%2FTwitter
```

Both systems have persisted read/write grants with `sourceUserId=0`, `targetUserId=999`, and `targetPkg=com.twitter.android`. In other words, the clone is using a granted folder at `/storage/emulated/0/Download/Twitter`.

### Remaining difference, and why it is acceptable to me

The system folder picker no longer offers the separate user-999 storage on either phone. That storage still exists. However, the standard DocumentsUI and ExternalStorageProvider packages are installed only for user 0, and both clone profiles report `mMediaSharedWithParent=true`.

This is consistent with a limitation or policy in the system picker/profile integration. It does not establish whether it is an OEM bug, an intentional sharing policy, or behavior common to all Android clone implementations. The user-visible tradeoff is clear: the new route saves into owner storage, and I currently cannot choose the clone's separate storage through the picker.

That is acceptable for my current use: reliable downloads into an explicitly selected, accessible folder satisfy the practical need. It is a successful replacement for the old workflow, with different storage isolation, rather than proof that the old DownloadManager-to-user-999 route has itself been repaired.

Native X/premium/offline downloads are outside this update's verification scope. I am also not claiming an exhaustive media-scanner/gallery refresh regression test; the read-only checks establish the actual destinations and grants, while the usability conclusion is my current user feedback.

**Please feel free to close this PR and #1866 as superseded.** I am leaving them open for the maintainer's decision. The unavailable user-999 storage choice is documented here as a remaining, acceptable limitation, not a request to merge the old implementation unchanged.

crimera:
ok, so am i right to assume that your concerns were solved with the refactor? i closed both pull requests.

Issue #1813, opened 2026-09-02
bug(Twitter): Native Downloader not working


GrandMaesterD:
### What happened?

for the past 5 patches or so the native downloader hasn't been working for me, I've tried wiping data/cache, fully reinstalling and changing the download folder
Whenever I try to download any image or video nothing I get the "Download started" popup but nothing gets downloaded and I have no notification about it

### Twitter/X version

v12.19.1-release.0

### Piko patch version

3.9.0

### Logs

```shell

```

### Confirm

- [x] I searched open and closed issues; this is not a duplicate.
- [x] I included all relevant details.

yusefgharib:
Same issue here

crimera:
were you using a vpn when downloading?

kyoyacchi:
> were you using a vpn when downloading?

I don't use vpn, nor dns. Just popup appears but nothing happens. One UI 8.5 Samsung

crimera:
btw what android version are you guys using?

kyoyacchi:
> btw what android version are you guys using?

I am using Android 16 idk about OP though. This was happening also in Komikku/Mihon forks they fixed it somehow. Sometimes even reddit doesn't download image/videos

GrandMaesterD:
> btw what android version are you guys using?

Android 16, no vpn

Issue #1787, opened 2026-08-29
Files Permission for downloads


ghost:
Is their any solution for this? I can't able to download any media
Device: poco x7 Android 16

<img width="1220" height="2712" alt="Image" src="https://github.com/user-attachments/assets/14dfcb63-a13f-455b-83b0-ef63d7e985f5" />

naoverse06:
create new folder

ghost:
> create new folder

I tried it already but still same

Xiaorui-Huang:
I unfortunately have the same problem, no matter how many create new folder I do, or nested ones, it does not allow me to use this folder. 

Usually on the navigation sidebar there is a downloads, just like in the Move app, that allows me to select folders in downloads. But that's not the case for PickowInstaGram, at least for me.

So I can't download anything without choosing a downloads folder.

There might be some bugs in the permission request or the folder search.

Issue #1641, opened 2026-08-02
[IG] cannot dowload posts when using dual app on xiaomi HyperOS


Billyyyy7:
When using Piko IG as a dual app on Hyper OS 2, I am unable to download posts.
I am prompted to select a download directory every time, and the download fails.
Presumably, as a dual app, a separate directory is created, so I should select the directory intended for the dual app; however, as the main directory appears in the directory selection screen for Piko IG, I suspect this is a bug caused by the inability to select the directory intended for the dual app.

Patch version. 3.8.0

mian196:
I don't know if its related to piko patches or the HyperOS dual app setup, you can use both accounts on same apo or use one in Instagram and one in piko ig using clone feature, or you can always install apps in second space or second user instead of main and all it takes is 1 click to switch

Billyyyy7:
I’ve been able to save files without any issues on other patched versions of Instagram, and as it’s possible to select a directory for dual apps, I don’t think this is a problem specific to Piko.
On ‘My Insta’ and the original Instagram, which I’ve set up as dual apps, I can save files to the dual app directory (similar to the Second Space directory).

Billyyyy7:
The attached image shows the download locations in the Dual App’s ‘My Insta’ and ‘Insta Eclipse’ sections. In the Dual App, a directory named ‘emulated/999’ is added.
Piko cannot select that directory

<img width="1220" height="171" alt="Image" src="https://github.com/user-attachments/assets/f7549dae-4600-445a-be41-8a31fcb70f4d" />
<img width="1220" height="211" alt="Image" src="https://github.com/user-attachments/assets/6be4c9ed-ec8c-4f96-aa91-7d300eedb7d9" />

PR #1866, opened 2026-09-11
fix(newx): support public media downloads in clone profiles

WayneShao:
## Problem

On OEM system clone profiles, X keeps the package name `com.twitter.android` but runs under Android user 999 with a separate storage namespace. The DownloadManager provider is installed for user 0, so public clone downloads can fail before bytes are written or published even when the owner profile works.

## Change

Add an opt-in NewX compatibility patch for public media downloads in the active OnePlus/OPPO/Xiaomi clone profile (user 999, API 29+). The normal user-0 DownloadManager path, private/offline requests and unrelated profiles remain unchanged.

The compatibility path:

- transfers public media in the clone app process;
- validates nonempty and complete responses before publication;
- writes and publishes through the current user's MediaStore;
- preserves callback handling through patch-time-resolved references;
- handles pending-name conflicts and duplicate requests;
- requests a single-file media scan after publication;
- retries `name=4096x4096` once when the exact X image `name=orig` endpoint returns 404, matching the existing upstream inline fallback.

## Latest target validation

This branch has been rebased onto upstream `x-lite` commit `68170cd326aeb1c3c5e39b3fc66468dda5782d73` and includes the latest local media-refresh prerequisite. Target input: X `12.25.2-prod.01`, versionCode `312252001`, APKM SHA-256 `e07e7bbb46a67b01872545968db6ba28db9cc1d87ac6b5f2fd5da2d138ea70f2`.

- 39 selected patches applied; `failedPatches=[]`.
- 209 extension tests and 40 patch tests passed.
- NewX resolver lint passed.
- Added application-link audit: 3 introduced app references checked, 0 unresolved.
- The signed resulting package was force-overwrite installed with `adb install -r --no-incremental` on Xiaomi HyperOS 4 / Android 17 and OnePlus ColorOS 16 / Android 16. Both user 0 and user 999 remain installed and user data was retained.
- Earlier combined builds completed real clone downloads and media-scan callbacks on both devices. The previously failing images on this exact revision remain pending user retest.

## Draft follow-up

- [ ] Retest the previously failing images on both devices and verify main-profile controls.
- [ ] Validate inline/clone patch-selection combinations and representative older/current/alpha APKs.
- [ ] Add Android integration coverage for MediaStore conflict/cleanup behavior, publication exceptions and native callback completion.
- [ ] Agree on lifecycle behavior after the host process is killed; persistent recovery and orphaned pending-item cleanup are not implemented.

The implementation is intentionally presented as a clone-profile compatibility option rather than DownloadManager lifecycle equivalence. Hidden existing files can receive provider rename suffixes, and cross-owner overwrite remains permission-constrained. Image-format preservation is separate from this patch.

Related reports: #1641, #1813 and #1787 are referenced for context only; their root causes have not been established as identical and this PR does not automatically close them.


crimera:
hello, what's a cloned profile?

WayneShao:
Thanks for asking. By “cloned profile” I mean the OEM system-level Dual Apps / App Cloner instance, not a second physical Android user and not a separate package name.

I verified this on the two affected phones on 2026-09-14:

- Xiaomi 17 Ultra for Leica (`nezha`, HyperOS 4 / Android 17): the system creates Android user `999` named `XSpace`, with `Type: android.os.usertype.profile.CLONE`. X is installed for both user 0 and user 999. The package UID is `11000` for user 0 and `99911000` for user 999.
- OnePlus 13T (`PKX110`, ColorOS 16 / Android 16): the system creates Android user `999` named `MultiApp`, also with `Type: android.os.usertype.profile.CLONE`. X is installed for both user 0 and user 999. The package UID is `11898` for user 0 and `99911898` for user 999.

The package name remains `com.twitter.android` in both profiles. Android distinguishes the clone through the user/profile ID and UID. The two storage namespaces are therefore separate:

```text
/storage/emulated/0/       main profile
/storage/emulated/999/     OEM clone profile
```

For example, a clone download is expected under a path such as `/storage/emulated/999/Movies/Twitter/` or `/storage/emulated/999/Pictures/Twitter/`, while the main profile uses the corresponding `/storage/emulated/0/` path.

The relevant difference for this issue is the system download provider. On both phones, `com.android.providers.downloads` is installed for user 0. There is no separate provider package installed for user 999. Clone requests therefore cross the profile boundary into the owner process while retaining the clone caller identity and the user-999 storage destination. The main profile and clone profile do not use equivalent caller/storage routing even though the application package name is the same.

That is the reason this patch is scoped as an opt-in compatibility patch for the clone profile. It does not replace the normal user-0 DownloadManager path. When the patch is not selected, the inline downloader keeps its original routing. When selected and running as user 999 on the affected OEM/API range, public media is transferred in the X clone process and then published through MediaStore for that same user. Private/offline requests retain their native path.

This also explains why the symptom can be selective: the main profile can download normally, some clone media requests can succeed, while other clone requests fail before bytes are written. The failure is at the cross-profile download/storage boundary, not evidence that the media URL itself is invalid. In our Xiaomi logs, some image requests returned HTTP 404 in the initial app-process transport while other images completed; the same previously failing images were downloadable through another X plugin, which is why the patch now preserves the upstream image-size fallback as well.

The current patch is deliberately disabled by default and has explicit runtime gating for the clone profile. The PR description also documents the remaining limits: no persistent recovery after the host process is killed, no claim of identical DownloadManager background lifecycle, and pending Android integration tests for all patch-selection combinations.

WayneShao:
One broader clarification: this is not a Xiaomi- or OnePlus-specific concept. System-level Dual Apps / App Cloner is a common feature across many Chinese Android OEMs, including Xiaomi/HyperOS, OPPO/ColorOS, OnePlus/ColorOS, vivo, Huawei and others. The implementation is OEM-specific, but the usual model is an Android clone/profile user separated from the owner user, with `/storage/emulated/999/` commonly used for the clone storage namespace.

User `999` is a widely used convention for these OEM clone profiles (and is what both affected phones use), although it is not an Android API guarantee and should not be treated as universal for every vendor, ROM or device generation. The patch therefore uses the profile/user boundary plus explicit OEM/API conditions, rather than relying on the package name alone or claiming that every device with a second profile has the same DownloadManager behavior.

WayneShao:
Update after moving to current upstream `x-lite`: I consider the practical problem that motivated this PR addressed for my current use case, and I am comfortable with this PR being closed as superseded. There is an important storage-behavior difference to record.

**Original problem:** the main X instance in user 0 downloaded normally, while the OEM clone in user 999 encountered download-start/permission failures with the DownloadManager-based flow. The system DownloadProvider was installed for user 0, not separately for user 999; that was evidence of a different environment, not by itself proof of every failed request's exact cross-user routing. Successful downloads in the previous clone-storage setup lived under `/storage/emulated/999/...`, with visibility through the OEM gallery/file-manager integration. Some completed files also took time to appear in the gallery.

**What our PRs tried to achieve:** reliable public image/video downloads from the clone, correct publication and media discovery, without changing the main instance's normal route. This PR used app-process HTTP transfers and the clone user's MediaStore, with response validation, failure cleanup, name-conflict handling, and the existing image-size fallback. #1865 added single-file scanning/notification after publication to help media discovery.

**Why the new version is sufficient for me:** both phones now run X `12.28.0-prod.01`, patched from `x-lite` at `0e30ed414f459dbbd3fb73227936d5db89662908`, without either local PR. The new inline-download flow uses its own transfer and a SAF-authorized destination, avoiding the old DownloadManager/staging/publication bridge. My current usage assessment is that this addresses the practical download problem.

**The remaining difference:** on both Xiaomi/HyperOS 4 and OnePlus/ColorOS 16, the system picker does not offer the separate user-999 storage. Read-only checks confirm both clones' image/video folders point to `/storage/emulated/0/Download/Twitter`, through a `content://0@com.android.externalstorage.documents/tree/...` URI, with persisted read/write permission granted from source user 0 to X in target user 999. The user-999 backing directory still exists, but the standard DocumentsUI/ExternalStorageProvider packages are installed only for user 0, and both clone profiles have `mMediaSharedWithParent=true`.

This suggests a system picker/profile-integration limitation or policy; it does not establish a universal Android bug. The new flow solves my practical need by saving reliably into the selected owner-profile folder, while separate clone-storage selection/isolation remains unavailable through that picker. I accept that tradeoff. The old user-999 DownloadManager route itself has not been demonstrated fixed, and native X/premium/offline paths and exhaustive gallery-refresh behavior are outside this update's verification scope.

The fuller explanation is also posted on #1865 in response to the maintainer's SAF-refactor question. **Please feel free to close this PR as superseded.** I am leaving it open for the maintainer's decision and preserving this remaining limitation in the report.

PR #1875, opened 2026-09-12
fix(Instagram): Sync ghost mode icons when settings change

naoverse06:


imPratyoosh:
What icon is used to show ghost mode?
I mean Myinsta uses this and it's so good.

naoverse06:
I’ll change the category icon to that one, but I don’t plan to change the eye icon in the action bar for now
https://github.com/tabler/tabler-icons

PR #1877, opened 2026-09-12
fix(Instagram): Skip event dispatch when analytics are disabled

WZSE:
Throwing an `IOException` for `/logging_client_events` in `Links.interceptUri()` caused Meta's native `fflogger_worker` thread to treat blocked requests as transient connection failures, triggering an
unthrottled background retry storm that consumed excessive CPU.

This bypasses event dispatch directly inside `EventBuilder.commit()` when `disableAnalytics` is enabled so events never enter the native logger queue. Also removes `/logging_client_events` from `Links.
interceptUri()` while keeping Graph API endpoints blocked.

## Testing
- `.\gradlew.bat :patches:checkStringResources --no-daemon --console=plain`
- Tested on Instagram v435.0.0.37.76 with Piko v3.10.0-dev.6
- Tested on Infinix X6838 (Android 14)


naoverse06:
>Tested on Instagram v435.0.0.37.76

You mean v439.0.0.37.89?

WZSE:
> > Tested on Instagram v435.0.0.37.76
> 
> You mean v439.0.0.37.89?

Actually no, but I realized and tested it with 439.0.0.37.89 as well now. Working fine.

PR #1884, opened 2026-09-13
feat(Instagram): Add story seen button

naoverse06:
Adds an eye button to the story header when "View stories anonymously" is enabled. tap it to mark the current story as seen

https://github.com/user-attachments/assets/0820ff54-1e7c-46fc-ad88-049d544b106b


vippium:
Hii @naoverse06 , there is an bug with the story seen button. It gets automatically hidden in the initial story. But when changing/swiping to different one, then it gets permanently appears. But it won't appear on that specific a/c from where we started seeing stories. 

Also make an enhancement in the toggle. 

1. Make the button only appear if user has enabled anonymous story viewing. If not, the button will not be shown, ad it'll be useless.
2. Currently, if the story is marked as seen, the button doesn't is greyed out. But if we have watched the story previously or marked it as seen, then it'll be greyout. 

So, it should work like after clicking on the eye, the button will be greyout out and a pop-up/toast will appear saying story marked as seen.

PR #1891, opened 2026-09-14
feat(Instagram): Customize the text color and thought bubble of the notes.

ViictorNunes:
It adds a new button next to the profile when creating a note; clicking it opens a window that lets you customize the note's text color from a variety of options. Once you enter the HEX code (you can use a Google color picker to make this easier) and confirm, you are automatically taken to the theme selection window—unfortunately, there is no way to bypass this step without selecting a theme (I’ll look into that later). After you choose a theme, the color is applied and can be used for songs and other items. And yes, it is visible to everyone.

https://github.com/user-attachments/assets/2a4edf9f-d4a4-4c8f-a21d-f4aaffc7ed7f


swakwork:
If the color text field is custom made then, it would be great if it was a color keeper rather than a text input. Users can also easily add the color.

Also for the icon go for a ink dropper symbol instead of `Aa`

swakwork:
Note to who ever who is going to merge it, Squash and merge please

ViictorNunes:
What do you think? I went ahead and added an option to color the thought bubble. I'm still going to make some changes

https://github.com/user-attachments/assets/65747ae5-2d1c-4892-a94d-513cb7535646

vippium:
Won't it trigger any issue from Meta ?? 

Also, rahter than putting them simultaneously. We can arrange them in vertically. just in the right side. What's your opinion ??

ViictorNunes:
> Won't it trigger any issue from Meta ??

no. has other resources that could give some problem that until today have not done 

> Also, rahter than putting them simultaneously. We can arrange them in vertically. just in the right side. What's your opinion ??

I thought about it, but I would have to leave it small and I was conflicting the color button. 


I am making a few more changes and until then I will put how it will look in the final result.

PR #1896, opened 2026-09-15
Add Hide Chat feature: hide a DM conversation from the inbox via a pr…

red2kit:
Adds a "Hide this chat" button on a user's profile page. When tapped, it stores the person's username/full name and hides their conversation row from the DM inbox list — while keeping the chat itself fully accessible if you still have the link/search. Includes a settings toggle ("Enable hidden chats") and a way to view/restore hidden chats from Piko settings.


vippium:
Greatt ... If possible try to make it like locked chats. Which'll be hidden automatically but ask for passcode or biometric when accessing them.

naoverse06:
1. Open a PR against the dev branch
2. Do not modify any strings other than the English ones

